PR #2, opened 2015-03-16
http extensions - fix charset

mwallnoefer:
- in the HTTP extension we have this "charset" parameter, which at the moment we enforce for outgoing messages (perfectly correct) but also for incoming ones if no other charset has been specified.
- The latter behaviour is in clear contradiction to HTTP's RFC (http://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html#sec3.4.1) which enforces the use of latin1 (iso-8859-1) in that case.
- I got across over this limitation only when setting our charset parameter testwise to "utf-16". The content gets delivered correctly, but webbrowsers like Firefox continue to send POST requests (form data in urlencoded) in URL-encoded ASCII (in theory also latin1 which is the superset).

### DIFF
--- a/extensions/http/src/jolie/net/HttpProtocol.java
+++ b/extensions/http/src/jolie/net/HttpProtocol.java
@@ -841,7 +841,7 @@ public class HttpProtocol extends CommProtocol
 		throws IOException
 	{
 		Method method = send_getRequestMethod( message );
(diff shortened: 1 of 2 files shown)


fmontesi:
I've read the specs. They're a bit convolute but I agree. Merging!

PR #7, opened 2015-03-19
FileService - add raw to string converters with charset enforcement

mwallnoefer:


fmontesi:
I understand why you put them there (legacy, we had a similar method there), but I disagree with the initial decision of putting them there.

I would make it so the existing conversion methods in File print a "deprecated" warning message on the Interpreter logger, and create a separate Java service for this kind of conversions.

What do you think?

fmontesi:
PS: great that we have the implementation for more conversion methods. :-)

fmontesi:
Very nice now. :-)

How about we call the service simply "Converters" for simplicity?

Also, I'd say, e.g., rawToBase64 rather than fromRawToBase64

- F

PR #29, opened 2015-04-12
jolie2java improvements

mwallnoefer:
I have reworked jolie2java and added more automation regarding type conversions. This was necessary since the code lacked the long and raw data types. Fortunately we can drop the various "else"s since we are always precisely checking for the correct type: isBool(), isDouble(), instanceof Boolean, instanceof Double.


fmontesi:
Looks good, I'm pulling it in for Jolie 1.3.

PR #30, opened 2015-04-12
GWT - drop gwt-dev.jar which we do not provide any longer

mwallnoefer:
This version is outdated and generally it is too big (last versions ~ 37 MB)
to be provided within Jolie.


fmontesi:
Excellent. Let's wait for this until Jolie 1.3 though, as it is a big change. Also I'd like to test it against the client app by @klag as soon as he can send it in issue #30

Issue #27, opened 2015-04-09
Internal services


fmontesi:
Hi all,

I'm reposting this from my original e-mail to the mailing list for Jolie 1.2. I've postponed this feature due to lack of discussion, so I'd like to rekindle that here.

# Motivation

Right now doing procedural programming in Jolie is a pain, as we do not support local variables in defined procedures. This is *intentionally* painful, as using procedures too much can make it hard to switch to a distributed implementation based on communications later on.

A procedure basically acts as a request-response invocation: we call it, and then we wait until it's done. Therefore we usually implement recursive algorithms by having a service call itself on one of its operations (e.g., [see here](http://thesave.github.io/jolie/microservices/tutorials/2015/02/03/Implementing%20QuickSort-as-a-MicroService.html)). This is a bit cumbersome sometimes and promotes bad design (algorithms should be implemented in a separate service most of the time, but creating separate services is a bother sometimes). So we need syntax sugar! All the elements are basically there, we only need to find a nice syntax and do some plumbing in the interpreter to make it work by using the mechanisms we already have. So what I would really like to have is the possibility of defining lightweight "internal" services inside of a Jolie program.

# Requirements

* Internal services should be easy to code, as in quicker than coding a fully-fledged Jolie service. We want syntax sugar. The whole reason for introducing them is to avoid the bother of defining a separate Jolie file to embed, with its interface and ports everything.
* We should also be able to have many internal services with different names, so that the programmer is left free to split operations as she sees fit.
* Finally, internal services should have the same lifecycle as a normal embedded service.

# Initial Proposal

We introduce a keyword, e.g., "service", to define internal services that works like this:

<pre>
interface SrvIface {
OneWay: op(void)
RequestResponse: op2(T)(T2)
}

service Srv {
Interfaces: SrvIface
Code: {
    [ op1() ] { ... }
    [ op2(x)(y) { ... } ]
}
}
</pre>

Srv would work as an embedded service, so this would get compiled to:

<pre>
outputPort Srv {
Interfaces: SrvIface
}

embedded {
Jolie: *internal reference to the code of the Srv* in Srv
}
</pre>

So that now we can use Srv inside of our Jolie program:

<pre>
main {
    op2@Srv( x )( y )
}
</pre>

# Discussion points

* Should Srv inherit anything from the enclosing service? Some examples follow.
* Can Srv use the output ports of the main service?
[...]

PR #45, opened 2015-05-18
Internal services

mawolf:
Adding support for internal services


jolie/src/jolie/Interpreter.java
```diff
@@ -1182,26 +1213,33 @@ private boolean buildOOIT()
 				} else {
 					throw new InterpreterException( "Input compiled program is not a JOLIE program" );
 				}
-			} else {
+			}
+            else if (this.internalServiceProgram != null) {
+                program = this.internalServiceProgram;
+                OLParseTreeOptimizer optimizer = new OLParseTreeOptimizer( program );
```
fmontesi: Looks like this code is duplicated with the one in the next branch.

jolie/src/jolie/runtime/embedding/EmbeddedServiceLoader.java
```diff
@@ -100,4 +107,77 @@ protected void setChannel( CommChannel channel )
...
+    public static abstract class EmbeddedServiceConfiguration
+    {
+        private final Constants.EmbeddedServiceType type;
+        
+        public EmbeddedServiceConfiguration(Constants.EmbeddedServiceType type)
+        {
+            this.type = type;
+        }
+        
+        public Constants.EmbeddedServiceType getType()
```
fmontesi: Please rename getType() to type()

jolie/src/jolie/runtime/embedding/EmbeddedServiceLoader.java
```diff
@@ -100,4 +107,77 @@ protected void setChannel( CommChannel channel )
...
+         *
+         * @param type Type of embedded service, cannot be INTERNAL
+         * @param servicePath path of service
+         */
+        public ExternalEmbeddedServiceConfiguration( Constants.EmbeddedServiceType type, String servicePath )
+        {
+            super(type);
+            this.servicePath = servicePath;
+            
+            assert type != Constants.EmbeddedServiceType.INTERNAL;
```
fmontesi: Should probably throw IllegalArgumentException in case type is INTERNAL

libjolie/src/jolie/lang/parse/OLParser.java
```diff
@@ -811,8 +812,160 @@ private void parseInterfaceOrPort()
...
+        InputPortInfo iport = null;
+        try {
+            iport = new InputPortInfo(
+                getContext(),
+                serviceName + "InputPort", //input port name
+                new URI( Constants.LOCAL_LOCATION_KEYWORD ),
+                null,
+                protocolConfiguration,
+                new InputPortInfo.AggregationItemInfo[] {},
+                new HashMap<String, String>() );
```
fmontesi: Use Collections.emptyMap()

libjolie/src/jolie/lang/parse/OLParser.java
```diff
@@ -811,8 +812,160 @@ private void parseInterfaceOrPort()
...
+        String serviceName = token.content();
+        
+        //validate token
+        getToken();
+        eat(Scanner.TokenType.LCURLY, "{ expected");
+        
+        //initialize internal interface and interface list
+        List< InterfaceDefinition > interfaceList = new ArrayList< InterfaceDefinition >();
+        
+        OLSyntaxNode internal_main = null;
```
fmontesi: internal_main -> internalMain

libjolie/src/jolie/lang/parse/OLParser.java
```diff
@@ -811,8 +812,160 @@ private void parseInterfaceOrPort()
...
+        SequenceStatement internal_init = null;
```
fmontesi: internal_init -> internalInit

fmontesi: internal_init could probably just be OLSyntaxNode, SequenceStatement seems redundant.

libjolie/src/jolie/lang/parse/util/VisualizeAST.java
```diff
@@ -0,0 +1,458 @@
+/***************************************************************************
```
fmontesi: Where is this class used? Please remove it if unused (we could move it to an external library though maybe... but let's leave it for another patch).

libjolie/src/jolie/lang/parse/OLParser.java
```diff
@@ -811,8 +812,160 @@ private void parseInterfaceOrPort()
...
+            throwException( "You must specify a main for internal service " + serviceName );
+      
+        Program internalServiceProgram = new Program(getContext());
+        
+        // copy children of parent to embedded service
+        for ( OLSyntaxNode child : program.children() ) {
+            if ( child instanceof TypeInlineDefinition ||
+                child instanceof InterfaceDefinition ||
+                child instanceof OutputPortInfo ||
+                child instanceof EmbeddedServiceNode )
```
fmontesi: This does not look like the right way of implementing the first requirement in Semantics of #27 .
It's going to duplicate embedded services, which could be a problem in case of sub-services written in Jolie (and even Java, although many are harmless, like Console).

What we should do is implement some mechanism for the internal service to wait for the bindings for the output ports at runtime. This doesn't look too easy, we need to think of the lifecycle defined in:

https://github.com/jolie/jolie/blob/master/jolie/src/jolie/process/InitDefinitionProcess.java

Embedded services are loaded before the init procedure starts, so the internal service can't wait for the parent's init to end before "copying over" the bindings for the output ports or we'll create a deadlock.

fmontesi:
Great patch, thanks. :-)

Indentation looks weird sometimes. Please follow the style of the rest of the code (spaces in parentheses) and use tabs instead of spaces.

@jolie/developers More eyeballs on the patch are welcome.

fmontesi:
The formatting rules I used (updated to last year I think, but netbeans should not have changed so much hopefully?..)

http://filebin.ca/22Ay89ym7QDR/formatting.zip

mawolf:
Again, thank you for the comments and recommendations!

The latest commits should fix the concerns that you had. Let me know if further changes are needed!

PR #57, opened 2015-08-27
choice operator implementing (for processing types like "type choice: void {.x: void { .y: int | string}}")

agwe:
Changes in Type cast/check and castSubType, checkSubType methods
Adding new classes TypePair (for making cast/check operations over two types) and TypeChoiceDefinition (to store structure with two types)
Adding corresponding visitors to OLParser, OLParserTreeOptimizer, OLVisitor, SemanticVerifier, ProgramInspectorCreatorVisitor


fmontesi:
Closing, looks like an old pull request.

PR #61, opened 2015-09-28
clean old typeDefinition methods from Jolie2Dummy tool

agwe:
I have optimized a bit convertTypes method. Please tell me, if the changes are not ok.
Now It returns dummy reposnse for both types of typeChoiceDefinition type. Is that ok, or should work differently? 

I have also faced the problem that inputPortInfo.context().source().getSchemeSpecificPart().contains( directorySourceFile ) never worked on my windows machine, because of different string format:
"/C:/jolie_examples/server.ol" comparing to "C:\jolie_examples\".

### DIFF
--- a/tools/jolie2dummy/src/joliex/dummycreator/impl/JolieDummyDocumentCreator.java
+++ b/tools/jolie2dummy/src/joliex/dummycreator/impl/JolieDummyDocumentCreator.java
@@ -34,7 +34,9 @@ import jolie.lang.parse.ast.InterfaceDefinition;
 import jolie.lang.parse.ast.OperationDeclaration;
 import jolie.lang.parse.ast.RequestResponseOperationDeclaration;
 import jolie.lang.parse.ast.types.TypeDefinition;
+import jolie.lang.parse.ast.types.TypeInlineDefinition;
 import jolie.lang.parse.ast.types.TypeDefinitionLink;
+import jolie.lang.parse.ast.types.TypeChoiceDefinition;
 import jolie.lang.parse.util.ProgramInspector;
 
 /**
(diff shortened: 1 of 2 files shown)


tools/jolie2dummy/src/joliex/dummycreator/impl/JolieDummyDocumentCreator.java
```diff
@@ -120,239 +122,51 @@ private void convertOperation( OperationDeclaration operatationDeclaration, Stri
...
-					}
-					stringBuilder.append( "response<<dummyResponse\n" );
-				}
-
-			}
-
-			/// handeling of types here
-
+			convertTypes( typeDefinition, stringBuilder, "dummyResponse");
+			stringBuilder.append("response<<dummyResponse\n" );
```
fmontesi: Indentation (space before the string parameter).

tools/jolie2dummy/src/joliex/dummycreator/impl/JolieDummyDocumentCreator.java
```diff
@@ -120,239 +122,51 @@ private void convertOperation( OperationDeclaration operatationDeclaration, Stri
...
-
-					}
-					stringBuilder.append( "response<<dummyResponse\n" );
-				}
-
-			}
-
-			/// handeling of types here
-
+			convertTypes( typeDefinition, stringBuilder, "dummyResponse");
```
fmontesi: Indent (space after last parameter).

tools/jolie2dummy/src/joliex/dummycreator/impl/JolieDummyDocumentCreator.java
```diff
@@ -120,239 +122,51 @@ private void convertOperation( OperationDeclaration operatationDeclaration, Stri
...
 	}
 
-	private void convertTypes( TypeDefinition typeDefinition, StringBuilder stringBuilder, String nameVariable )
-	{
-
-
-		if ( typeDefinition instanceof TypeDefinitionLink ) {
-
+	private void convertTypes( TypeDefinition typeDefinition, StringBuilder stringBuilder, String nameVariable) {
+		if (typeDefinition instanceof TypeChoiceDefinition) {
```
fmontesi: Indent: space after "(" and space before ")"

(to make it consistent with the rest of the codebase. same rule for the rest)

tools/jolie2dummy/src/joliex/dummycreator/impl/JolieDummyDocumentCreator.java
```diff
@@ -120,239 +122,51 @@ private void convertOperation( OperationDeclaration operatationDeclaration, Stri
...
+			TypeChoiceDefinition typeChoiceDefinition = (TypeChoiceDefinition) typeDefinition;
+			convertTypes(typeChoiceDefinition.left(), stringBuilder, nameVariable);
+			convertTypes(typeChoiceDefinition.right(), stringBuilder, nameVariable);
```
fmontesi: Mmh I think that this line should be erased. Assume that we have this type:

type t : int | string

Then your implementation would generate something like:

x = 42;
x = "dummy"

right? The second one is redundant, as it overwrites the first.

fmontesi:
Looks fine, I've added some minor comments. I'll merge it as soon as they are fixed!

PR #62, opened 2015-09-30
clean jolie2surface from old TypeDefinition methods

agwe:


tools/jolie2surface/src/joliex/surface/SurfaceCreator.java
```diff
@@ -169,73 +165,96 @@ private String getMax( int max )
 
 	private String getCardinality( Range card )
 	{
-		return "[" + card.min() + "," + getMax( card.max() ) + "]";
+		return ( card.min() == 1 && card.max() == 1 ) ? "" : ("[" + card.min() + "," + getMax(card.max()) + "]");
```
fmontesi: Good catch. :-)

tools/jolie2surface/src/joliex/surface/SurfaceCreator.java
```diff
@@ -169,73 +165,96 @@ private String getMax( int max )
...
 	}
 
+	boolean choice;
```
fmontesi: Please specify explicitly the scope (private?).

tools/jolie2surface/src/joliex/surface/SurfaceCreator.java
```diff
@@ -169,73 +165,96 @@ private String getMax( int max )
...
-                                   
-				}
+
+				if ( ( (TypeInlineDefinition) type ).untypedSubTypes() ) {
+					System.out.print( " { ? }" );}
+
 			}
-			types_vector.add( type.id() );
+		} else if ( type instanceof TypeChoiceDefinition ){
+			checkType( ( (TypeChoiceDefinition) type ).left(), 1 );
```
fmontesi: Does this way of using static 1 and 3 still work if left() or right() are yet again TypeChoiceDefinition?

fmontesi:
@klag this is the patch for jolie2surface to include choice types. Can you give a quick look that it doesn't break your workflow?

PR #65, opened 2015-10-19
Jolie2java

agwe:
Brought back hasSubtypes() and subTypes() methods to TypeDefinition class and overided them in descendants. 
Write getTypes() method in JavaDocumentCreator to get all possible choice native types.
Move TypeImpl, TypeLink and TypeChoice inside Type class and made them all static.


fmontesi:
Hi Larisa,

I don't understand why we can implement subTypes() in type choice, since, among other problems, the same subtype may appear with different types, for example:

type t: void { .x:int } | void { .x:string }

Why do you need to reintroduce subTypes?

PR #70, opened 2015-11-17
Regex

agwe:
@fmontesi Could you please look at this first prototype for regular expressions. It is able to deal with things like:
type a: string("[A-Z]*")
type b: void {.x: string("[A-Z]*")}
type d: a | c

I thought it would be nice to do something for linked types as well, like "type e: a("[a-zA-Z]*")" but found that comparing regular expressions inclusion is not so trivial task, but it can't be solved any way. What do you think, is it worth to do?

I also left place for implementing predicate logic in numeric types. Going to do it as far as you say, if this code is suitable.


jolie/src/jolie/OOITBuilder.java
```diff
@@ -639,20 +639,20 @@ public void visit( TypeInlineDefinition n )
 		boolean backupInsideType = insideType;
 		insideType = true;
 
-		if ( n.untypedSubTypes() ) {
-			currType = Type.create( n.nativeType(), n.cardinality(), true, null );
+		if (n.untypedSubTypes()) {
```
fmontesi: Please keep the original indentation. :-) (Also in the rest of the code.)

libjolie/src/jolie/lang/nativeTypes/NativeType.java
```diff
@@ -28,45 +28,80 @@
...
-
-	private final static Map< String, NativeType > idMap = new HashMap<>();
-
-	static {
-		for( NativeType type : NativeType.values() ) {
-			idMap.put( type.id(), type );
-		}
+public class NativeType {
+
+	String refinement = "";
```
fmontesi: I don't think that refinement should live here.

It should be a member of StringType, since for now we use it only there.

libjolie/src/jolie/lang/nativeTypes/NativeType.java
```diff
@@ -28,45 +28,80 @@
...
+	NativeTypeEnum type;
```
fmontesi: This should be private and final.

libjolie/src/jolie/lang/nativeTypes/NativeType.java
```diff
@@ -28,45 +28,80 @@
...
+	public NativeType(NativeTypeEnum type) {
+		this.type = type;
+	}
+
+	public NativeType(String refinement, NativeTypeEnum type) {
+		this.refinement = refinement;
+		this.type = type;
+	}
+
+	public NativeType() {
```
fmontesi: I don't think this constructor makes sense, since we always want a native type enum set. Can we remove it?

Then, from each inheriting class, you can just call super. I include a comment about how to do this in StringType.

libjolie/src/jolie/lang/nativeTypes/StringType.java
```diff
@@ -0,0 +1,16 @@
+package jolie.lang.nativeTypes;
+
+/**
+ * Created by lsafina on 13/02/16.
+ */
+public class StringType extends NativeType {
+
+    public StringType() {
+        type = NativeTypeEnum.STRING;
```
fmontesi: Should be super( NativeTypeEnum.STRING );

libjolie/src/jolie/lang/nativeTypes/NativeType.java
```diff
@@ -28,45 +28,80 @@
...
-	{
-		this.id = id;
+	public NativeTypeEnum getType() {
+		return type;
 	}
 
-	public String id()
-	{
-		return id;
+	public void setType(NativeTypeEnum type) {
```
fmontesi: What's the usage for this? Can't we remove it? The type should be set in the constructor.

libjolie/src/jolie/lang/nativeTypes/StringType.java
```diff
@@ -0,0 +1,16 @@
...
+    }
+
+    public StringType(String refinement) {
```
fmontesi: The refinement should probably be a Pattern and not a String.

libjolie/src/jolie/lang/parse/OLParser.java
```diff
@@ -262,17 +267,25 @@ private void parseTypes()
 	private TypeDefinition parseType( String typeName )
 			throws IOException, ParserException
 	{
-		TypeDefinition currentType;
+		TypeDefinition currentType = null;
```
fmontesi: Why adding the = null?

libjolie/src/jolie/lang/parse/OLParser.java
```diff
@@ -321,7 +334,7 @@ private TypeDefinition parseSubType(String id, Range cardinality)
 			throws IOException, ParserException
 	{
 		NativeType nativeType = readNativeType();
-		TypeDefinition subType;
+		TypeDefinition subType = null;
```
fmontesi: Why add the = null?

PR #80, opened 2016-05-10
Added Absolute path operations to File 

Ezbob:
The motivation for this enhancement is to provide operations in the File interface to convert relative file paths into absolute file paths, using the Java 7 `Path` class.

The operations does not check whether the files exists (this is a task for the user if they want to), so they can be used to create new file paths for files that does not exist.

### DIFF
--- a/include/file.iol
+++ b/include/file.iol
@@ -81,14 +81,14 @@ RequestResponse:
 	 * Can be used to construct an absolute path for new files that does not exist yet.
 	 * Throws a InvalidPathException fault if input is a relative path is not system recognized path.
 	 */
-	toAbsolutePath( string )( string ) throws JavaExceptionType( InvalidPathExceptionType ),
+	toAbsolutePath( string )( string ) throws InvalidPathExceptionType( JavaExceptionType ),
 
 	/**!
 	 * Constructs the path to the parent directory. 
 	 * Can be used to construct paths that does not exist so long as the path uses the system's filesystem path conventions.
 	 * Throws a InvalidPathException fault if input path is not a recognized system path or if the parent has no parent.
 	 */
-	getParentPath( string )( string ) throws JavaExceptionType( InvalidPathExceptionType ),
+	getParentPath( string )( string ) throws InvalidPathExceptionType( JavaExceptionType ),
 
 	/**!
 	  it returns if a filename is a directory or not. False if the file does not exist.
--- a/include/file.iol
+++ b/include/file.iol
@@ -79,14 +79,16 @@ RequestResponse:
 	/**!
 	 * Constructs an absolute path to the target file or directory. 
 	 * Can be used to construct an absolute path for new files that does not exist yet.
+	 * Throws a fault if input is a relative path is not system recognized path.
 	 */
-	getAbsolutePath( string )( string ),
+	toAbsolutePath( string )( string ),
 
 	/**!
-	 * Constructs an absolute path to the parent directory of the target file or directory.
-	 * Can be used to construct an absolute path to the parent directory for files that does not exists yet.
+	 * Constructs the path to the parent directory. 
+	 * Can be used to construct paths that does not exist so long as the path uses the system's filesystem path conventions.
+	 * Throws a fault if input path is not a recognized system path or if the parent has no parent.
 	 */
-	getAbsoluteParentPath( string )( string ),
+	getParentPath( string )( string ),
 
 	/**!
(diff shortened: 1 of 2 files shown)


include/file.iol
```diff
@@ -77,6 +77,18 @@ type ListResponse:void {
 interface FileInterface {
 RequestResponse:
 	/**!
+	 * Constructs a absolute path to the target file or directory.
```
fmontesi: a absolute -> an absolute

include/file.iol
```diff
@@ -77,6 +77,18 @@ type ListResponse:void {
 interface FileInterface {
 RequestResponse:
 	/**!
+	 * Constructs an absolute path to the target file or directory. 
+	 * Can be used to construct an absolute path for new files that does not exist yet.
+	 */
+	getAbsolutePath( string )( string ),
```
fmontesi: I'd probably call it toAbsolutePath, following Java in toString, toAbsolutePath, etc.

@jolie/developers Comments on the convention for transformer operations? I don't have *very* strong feelings here. (yet?) :-)

include/file.iol
```diff
@@ -77,6 +77,18 @@ type ListResponse:void {
...
+
+	/**!
+	 * Constructs an absolute path to the parent directory of the target file or directory.
+	 * Can be used to construct an absolute path to the parent directory for files that does not exists yet.
+	 */
+	getAbsoluteParentPath( string )( string ),
```
fmontesi: Why not simply getParentPath ? Then using this becomes a matter of:

toAbsolutePath@File( path )( path );
getParentPath@File( path )( path )

fmontesi:
Awesome, thanks. I couldn't finish my comments before, put the last ones now.

Ezbob:
Refactored the code a again, as suggested by @fmontesi . 

This time I also added faults; if the path is not recognized as a valid path by the file system then we get a fault, and in the case of `getParentPath` if the parent path cannot exist (i.e. get the parent of the root "/" on nix machines) then we also get a fault.

javaServices/coreJavaServices/src/joliex/io/FileService.java
```diff
@@ -744,4 +745,38 @@ public boolean accept( File directory, String filename )
...
+	@RequestResponse
+	public Value toAbsolutePath( Value request ) throws FaultException
+	{
+		Value response = Value.create();
+		String fileName = request.strValue();
+		
+		String absolutePath = Paths.get( fileName ).toAbsolutePath().normalize().toString();
+
+		if ( absolutePath == null ) {
+			throw new FaultException( "Not a valid path", new IOException() );
```
fmontesi: Should be

throw new FaultException( new IOException( "Not a valid path" ) );

Same for the other method.

fmontesi:
Thanks! I added a comment on a bug I noticed only now, and then we're good to go!

Ezbob:
Ahh... We got an unchecked exception that `Paths.get` throws... I'm just gonna wrap that up in a `FaultException`..

include/file.iol
```diff
@@ -77,6 +77,20 @@ type ListResponse:void {
 interface FileInterface {
 RequestResponse:
 	/**!
+	 * Constructs an absolute path to the target file or directory. 
+	 * Can be used to construct an absolute path for new files that does not exist yet.
+	 * Throws a fault if input is a relative path is not system recognized path.
+	 */
+	toAbsolutePath( string )( string ),
```
fmontesi: Should declare that it can throw InvalidPathException, with its type (probably JavaExceptionType?). Same for the one below.

include/file.iol
```diff
@@ -77,6 +77,20 @@ type ListResponse:void {
 interface FileInterface {
 RequestResponse:
 	/**!
+	 * Constructs an absolute path to the target file or directory. 
+	 * Can be used to construct an absolute path for new files that does not exist yet.
+	 * Throws a InvalidPathException fault if input is a relative path is not system recognized path.
+	 */
+	toAbsolutePath( string )( string ) throws JavaExceptionType( InvalidPathExceptionType ),
```
fmontesi: Huh? How does this work?

Should probably be: `throws InvalidPathException( JavaExceptionType )`

Same below.

fmontesi:
Oh, sorry, didn't notice that this was updated. I put in a last comment, promise. :-) I think that your interface would not compile now.

Ezbob:
Should be fixed now.

fmontesi:
Thanks. Merged. :+1:

Issue #84, opened 2016-06-15
Conflicting MatchRequest definitions for StringUtils and UriTemplates


D4vidH4mm3r:
Both string_utils.iol and uri_templates.iol define a type MatchRequest. The types are incompatible and so including both in a jolie program results in a SEVERE error (type MatchRequest has already been defined with a different structure).

I think one of the two types can simply be renamed and the problem should go away.

PR #85, opened 2016-06-15
Renaming MatchRequest to avoid conflict

D4vidH4mm3r:
As mentioned in #84, uri_templates.iol and string_utils.iol both define different types named MatchRequest, causing problems if both are included. In the small test I was working on, simply renaming the type in uri_templates.iol seems to work.


fmontesi:
Thanks! :+1:

Issue #51, opened 2015-06-08
In-range for loops


fmontesi:
# In-range for loops

## Problem

I'm getting sick of writing:

    for( i = 0, i < #myarray, i++ ) { println@Console( myarray[i] )() }

## Proposal

I propose:

    for( element -> myarray ) { println@Console( element )() }

Where element would be an alias to myarray[index] at each iteration.

## Discussion

Making element an alias to myarray[index] means that modifying the value of element inside of the loop would change the value of myarray[index]. In some language implementations, this is forbidden. However, it looks like this is pretty useful when manipulating structured documents, e.g., in XML. Opinions, @jolie/developers ?

## Another problem

However, aliases do not work with anonymous arrays, which worries me. For example, if we had a syntax such as `[ 1, 2, 3]`, this would not work:

    for( element -> [1, 2, 3] ) { code }

because aliases can only target paths, not anonymous data structures.

## Another solution

A safer (more future-proof) solution could be to have:

    for( element : myarray ) { code }

Where element is *copy* of myarray[index] at every iteration, but that would be very slow (remember that each Jolie var potentially has a big tree underneath).

A compromise I dislike but seems like the best option for now is:

    for( index : myarray ) { code }

where index is now instantiated from 0 to #myarray each time.

mwallnoefer:
Why not foreach?

foreach ( nameVar : root ) {
	//code block
}

fmontesi:
I hit submit too soon, see my updated text.

We could use `foreach` for the syntax, sure. I still have big doubts about the semantics.

fmontesi:
@mwallnoefer I just realised, did you mean encoding the array as subnode names?

mwallnoefer:
As I recall I just proposed to reuse the foreach keyword for case 1 (without copy) or case 2 (with copy).

fmontesi:
Dear @jolie/developers ,

This

```
foreach( element -> myarray ) { println@Console( element )() }
```

is now supported. Enjoy. :-)

thesave:
Wow, christmas has come early this year!
Very very very appreciated.
Thanks @braza501 :)

PR #87, opened 2016-07-12
added foreach loop for arrays v1 (Issue #51)

braza501:
According to Issue #51 (https://github.com/jolie/jolie/issues/51) I've implemented the first solution of the "foreach" syntax construct : for( element -> myarray ) { println@Console( element )() }  , although  the exact syntax is under discussion (e.g for_each or other name).


fmontesi:
Thanks! Looking into this right now. :-)

fmontesi:
@jolie/developers: I think I'm going to refactor this into:

foreach( item -> array ) { /* here item is an alias for the i-th element of array */ }

PR #89, opened 2016-07-14
Supporting rethrow

mawolf:
I'm a bit unsure if I also should do some checks, for example, if I'm in the main scope, a fault is thrown and the fault handler is rethrow.


fmontesi:
Awesome, thanks @mawolf! I'm away until mid August, I'll have a look then asap.

mawolf:
No worries! Have a great vacation! :-)

fmontesi:
I'm back! :-)

All looks good but one thing, which gives me a doubt. You take the name of the current fault from the local process state, but that can be changed by the programmer by using an assignment, which would yield a weird rethrow semantics. (For example, I could assign to the variable an arbitrary fault name and then `rethrow` would *not* rethrow the same fault.)

Maybe we can fetch the fault name from somewhere else?

mawolf:
To be honest I actually saw that as a feature(:D), however, I do see your point and will have another look. :-)

mawolf:
New change: I now use ExecutionThread to save the current fault. Is this in the direction you wish?

jolie/src/jolie/process/ReThrowProcess.java
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (C) 2016 Martin Wolf.
```
fmontesi: Do you have an e-mail to put here?

jolie/src/jolie/process/ReThrowProcess.java
```diff
@@ -0,0 +1,71 @@
...
+	public void run()
+		throws FaultException, ExitingException
+	{
+		if ( ExecutionThread.currentThread().isKilled() ) {
+			return;
+		}
+
+		ExecutionThread ethread = ExecutionThread.currentThread();
+		if ( expression == null ) {
+			throw ethread.currentFault();
```
fmontesi: What if currentFault() is null?

Which reminds me: we should probably insert a check in the semantic verifier that rethrow is used only from inside of something that is being installed using the `install` statement.

jolie/src/jolie/process/ReThrowProcess.java
```diff
@@ -0,0 +1,71 @@
...
+		} else {
+			throw new FaultException( ethread.currentFault().faultName(), expression.evaluate() );
```
fmontesi: What if currentFault() is null?

libjolie/src/jolie/lang/parse/OLVisitor.java
```diff
@@ -139,6 +140,7 @@
 	public void visit( InstallStatement n );
 	public void visit( CompensateStatement n );
 	public void visit( ThrowStatement n );
+	public void visit ( ReThrowStatement n );
```
fmontesi: There's an extra space before the left parenthesis.

libjolie/src/jolie/lang/parse/OLVisitor.java
```diff
@@ -177,5 +179,5 @@
 	public void visit( InlineTreeExpressionNode n );
 	public void visit( VoidExpressionNode n );
 	public void visit( ProvideUntilStatement n );
-	public void visit(TypeChoiceDefinition n);
+	public void visit( TypeChoiceDefinition n );
```
fmontesi: Cool, thanks. :-)

libjolie/src/jolie/lang/parse/Scanner.java
```diff
@@ -155,6 +156,7 @@
 		unreservedKeywords.put( "this", TokenType.THIS );
 		unreservedKeywords.put( "synchronized", TokenType.SYNCHRONIZED );
 		unreservedKeywords.put( "throw", TokenType.THROW );
+		unreservedKeywords.put( "rethrow", TokenType.RETHROW);
```
fmontesi: Lacking space before right parenthesis

libjolie/src/jolie/lang/parse/SemanticVerifier.java
```diff
@@ -386,6 +387,9 @@ public void validate()
 		}
 	}
 
+	@Override
+	public void visit( ReThrowStatement n ) {}
```
fmontesi: See my comment above. We should check that rethrow is used only if we're inside of an install.

libjolie/src/jolie/lang/parse/ast/ReThrowStatement.java
```diff
@@ -0,0 +1,55 @@
+/***************************************************************************
+ *   Copyright (C) by Martin Wolf                                          *
```
fmontesi: Use the nicer header you used before. And e-mail and year.

libjolie/src/jolie/lang/parse/ast/ReThrowStatement.java
```diff
@@ -0,0 +1,55 @@
...
+import jolie.lang.parse.context.ParsingContext;
+
+
+public class ReThrowStatement extends OLSyntaxNode
+{
+	private final OLSyntaxNode expression;
+
+	public ReThrowStatement( ParsingContext context )
+	{
+
```
fmontesi: Extra line?

test/primitives/rethrow.ol
```diff
@@ -0,0 +1,61 @@
...
+ *   Free Software Foundation, Inc.,                                       *
+ *   59 Temple Place - Suite 330, Boston, MA  02111-1307, USA.             *
+ *                                                                         *
+ *   For details about the authors of this software, see the AUTHORS file. *
+ ***************************************************************************/
+
+include "../AbstractTestUnit.iol"
+include "console.iol"
+
+define simpleReThrowTest
```
fmontesi: Indentation is inconsistent in this file.

fmontesi:
Fantastic, thanks! I think it's in good shape. I inserted detailed comments in the code. :-)

libjolie/src/jolie/lang/parse/SemanticVerifier.java
mawolf: Somehow it must be a mistake if a rethrow is executed outside of a fault handler, hence, if possible, the developer should be notified at compile time.

I think only throwing a RuntimeError at runtime is dangerous as the developer could find out too late about his mistake. The same could be said about just executing a no-op: this could result in the developer never discovering his fault and thereby risking that the service is put into an unwanted state.

fmontesi: Yeah, I agree. The only thing that wouldn't work is this:

```
define X
{
  rethrow
}

main
{
  install( default => X );
  throw( f )
}
```

which is fine but the semantic verifier would reject, I guess. Unless we're really really smart and we somehow check that when X is called it's always from inside of an install...! :-)

mawolf: Which also could be too restrictive. Consider the following example (not checked):
`define X
{
       if ( global.insideFaultHandler ) {
       	  rethrow
       }
}

main
{
     install( default => global.insideFaultHandler = true; X );
     throw( f ) 
}`

fmontesi: The only way to fix that is to define a primitive that checks whether we're doing fault recovery that we can statically determine.

But I don't see a good use of that. Why would you define a procedure that either recovers something or proceeds normally? You should just install something different. I generally dislike using bookkeeping variables since we have such a powerful mechanism as dynamic fault handling already. If anything, I would like people _not_ to use bookkeeping variables. But maybe there's something you cannot write without using them?

klag: Which is the problem? The define is just a macro and at the level of the language we should replace it everytime there is a call and then parsing the code. If at the moment the parser does not allow it, I think we can leave the runtime error when it is encountered outside an install.

fmontesi: No, that would break backwards compatibility. Right now you can write:

```
define X {
  if ( continue ) { X } else { nullProcess }
}
```

Which cannot be statically expanded during parsing (the expansion is infinite), it must be done at runtime.

mawolf: @fmontesi 
You are right! I have, without any luck, been trying to think of some cases where it actually would make sense to have a procedure that would be used in- and outside of a fault handler. However, the example would still be semantically correct as it wouldn't rethrow when called outside of the fault handler.

Another solution would be to print a warning at compile time whenever a procedure is referenced both inside and outside a fault handler, and then throw a RuntimeException if rethrow is executed outside.

fmontesi: Mmh yes. Slight variant:

- Throw a warning whenever a procedure that contains a rethrow is called from outside of a fault handler. Notice that being called from a handler for `this` does not count, since that's not a _fault_ handler.
- We need the runtime error anyway, to have a complete implementation of rethrow, so let's do it.

How about that? @jolie/developers this is basically the last bit, feedback?

fmontesi: @mawolf News about this?

mawolf: Sorry but no. Have been quite busy at work lately. :|

fmontesi: Hi @mawolf 

Any news on this one? :-)

- Fabrizio

mwallnoefer:
I think that this may be closed.

mwallnoefer:
Why hasn't this feature been merged? Have we implemented it in a different way?

PR #91, opened 2016-08-14
New doc

mawolf:
Support for /*\* _/, ///, and /_< */, //< comments.

Notice that I also changed the Scanner to use a BufferedReader for better performance on network storage and mechanical disks. Also, I have set the Scanner to ignore comments by default. It can be changed by giving a boolean in the constructor. This is for performance reasons.

What we still need to figure out is how we would want these new comments to be visualized in the joliedoc tool. I have changed it such that it outputs the subtype comments as... comments, just to show that the data is there. @thesave: do you have an opinion about this?

Please let me know if this is the direction that we wish to go.


klag:
Hi Martin, what about the support for /**!, is still available?

fmontesi:
I like it a lot so far. I'm not seeing how the buffered reader helps though. Many places that we call the scanner from, we already use a buffered input stream. So, if anything, I'd say that using a buffered reader should make performance _worse_. What's the catch?

mawolf:
@klag: you are correct. I will make it available again.

@fmontesi: you are also correct. :-) I did not see that the CommandLineParser  creates a buffered reader, only that the scanner would use InputStreamReader. Hmm, the weird thing is that I did some performance testing and saw that it actually was giving me better performance when parsing large files. Nevertheless, I will change it back! :-)

mawolf:
Latest commit should address those issues.

thesave:
Hi Martin, thanks for your work :)
Regarding your question, I put together a little demo [1] of a possible rendering for types. Let me know what you (all) think about it.

The link should be shared so that you can make changes and everyone with the link can see and modify it.

[1] https://jsfiddle.net/d7rg4jxh/#

mawolf:
Hi Saverio! The demo looks great! I find it clean an easy to understand. 

The only issue would be if the number of levels, in a type definition, would be large such that there would be no space left for the actual comment.

 If the rest also likes the design, I will start modifying joliedoc.

thesave:
I pondered a little on your point. From the top of my head, I would say that fixing a max-width percentage for the "type" column would do the trick. The best thing would be to have an example of what you refer to, so that we can see the final result and tweak the layout accordingly (maybe also confronting with the rendering of HttpConfiguration type [1] could give some insight).

Also, while programming in Jolie, I frequently go to the docs, copy-paste the type of a request and use it as a draft of the content that I want to send. Thus, it might be useful to keep a "clean" (stripped from comments) version of the type along with the structured annotation. Something along the lines of [2]. Note that I also added a div.annotation that wraps the annotated version of the type. With a little javascript (not necessarily in joliedoc but surely on the website) we can toggle annotations so that the documentation remains concise but expandible on demand.

[1] http://docs.jolie-lang.org/#!documentation/protocols/http.html
[2] https://jsfiddle.net/d7rg4jxh/2/

fmontesi:
If I like it? This is _awesome_ ! 8-)

@jolie/developers Have a look at Saverio's links to understand where this is going.

mawolf:
Is this output considered okay for file.iol?

https://jsfiddle.net/gq2pc25s/

klag:
Very very nice!!!!

Just a question because in this example I do not see the case: subnodes which declare other types are converted into html links? This is an example for clarify:

type T1: void

type T2: void {
   .node: T1
}

type T2 has a subnode which is defined as T1.

It could be very very very very very very useful to have a .node: <a href="#T1">T1</a> in  place of a simple string.

fmontesi:
This is very nice, agreed.

It find it a little redundant, for the types, to have both the code and then the description, the names of the subnodes is repeated and everything becomes pretty long. Can we merge them somehow?

thesave:
I tried to find some solution to merge the "code" and the described version of types but with no success.

As already mentioned a partial solution could be to use javascript to toggle the description (see a demo [here](https://jsfiddle.net/gq2pc25s/2/), do not look too much into the JS and CSS, in the final version we can definitely come up with something more elegant)

mawolf:
@klag: those will be linked as well.
@fmontesi: I completely agree. I was spending some time trying different things, however, I'm not sure how to avoid this redundancy. :-/
@thesave: I was wondering if we instead should hide the code snippet? However, I guess this depends on the type of user that reads the documentation.

fmontesi:
A cumbersome first attempt. See only `ReadFileRequest`:

thesave:
@fmontesi I can't spot the differences (I just looked at `ReadFileRequest`) wrt the one in my previous reply. What changed? :)

@mawolf you are right, it probably depends on what the user is looking for in the docs.

---

Btw, in my first demo I wrongly removed the root of the type. Imho we would be better off reporting the whole type (i.e., starting from the root and not from the first-level children) because 1) removing the root could be misleading, in particular to new users and 2) also the root can contain important comments.

For example
```java
type GetDateTimeRequest:long /*< the date in milliseconds */{ 
    .format?:string //< the date format in date and time pattern string
}
```
However, iirc (in case, correct me), this poses the problem of understanding when a comment before the root refers only to it (requiring the comment to be displayed next the root node) or to the whole type (which should be displayed before the description of the type). On the example below, we do not know if the programmer wanted to describe the whole type or just the root node.

```java
/** the date in milliseconds */
type GetDateTimeRequest:long { 
    .format?:string //< the date format in date and time pattern string
}
```

What do you think?

fmontesi:
Damn, I saved the wrong link. :-(

I had changed the description part to have also the root node, use more or less the same style (fonts, highlighting) of the type reported above, and report curly brackets. Then, the description of each node would appear on the right in a box with rounded corners.

fmontesi:
@thesave Good doubt... @jolie/developers opinions?

klag:
I think we should leave both the comment about the overall type, and the comment about the root node. 
Something like this

```
/** the date in milliseconds */
type GetDateTimeRequest:long   /// the date time in millisenconds
{ 
    .format?:string /// the date format in date and time pattern string
}
```

fmontesi:
I agree, but I think that the technical problem is that the two override the other in the AST? @thesave @mawolf , could you elaborate?

thesave:
As @klag wrote in the example, comments around the root of a type are ambiguous whether they apply to the root or to the whole type (in this second case the comment should be reported before the annotation of the single nodes.)

I gave it some thoughts. What about adding a javadoc-like tag (e.g., `@type`) to mark that the annotation describes the whole type and not a node?
Something like:
```java
/**@type The type for GetDateTime requests **/
/** the date in milliseconds */
type GetDateTimeRequest:long { 
    .format?:string //< @type .format is the patterns string of the response
                    //< the date format in date and time pattern string
}
```
As usual, the tag applies to the type pointed by the comment (i.e., /\*\* \*\*/ is a forward comment, /\*< \*/ is a backward comment) and can be stacked with other comments that refer to it. Hence, the doc of the example should look like this:

## GetDateTimeRequest
The type for GetDateTime requests
 .format is the patterns string of the response
<table><tr><td>type GetDateTimeRequest: long</td><td>the date in milliseconds</td></tr>
<tr><td>.format?:string</td><td>the date format in date and time pattern string</td>
</table>

---

Full disclosure: at first I also thought to use a `@node` tag for the nodes of types instead of the whole type, but I opted for the `@type` tag for less verbosity (types can be many but nodes of types can be many many :P)

mawolf:
Sorry for the delay!

@fmontesi: That is correct. In the current pull request, the comment value for the root node would be overwritten. However, I can change that rather quickly.

fmontesi:
@mawolf Looking forward to a description of how you'd do it. :-) @thesave That'd solve the problem, right?

thesave:
Yes it would be a solution to the problem however I would like to weight and discuss our options:

<ul><li> the `@type` tag would be relatively minimal for programmers (most likely one per type). However, it would require us to update the types in the standard library with the tag. 
Considering the example below
```Java
/**@type The type for GetDateTime requests **/
/** the date in milliseconds */
type GetDateTimeRequest:long { 
    .format?:string //< @type .format is the patterns string of the response
                    //< the date format in date and time pattern string
}
```
The tag would also permit to add general comments (i.e., referred to the whole type) next to nodes. Ideally, joliedoc should merge these additional comments in the general description of the type (following their order from top to bottom), like in example of rendering  below 
## GetDateTimeRequest
The type for GetDateTime requests
 .format is the patterns string of the response
<table><tr><td>type GetDateTimeRequest: long</td><td>the date in milliseconds</td></tr>
<tr><td>.format?:string</td><td>the date format in date and time pattern string</td>
</table>
</li>
<li> the `@node` is dual to the `@type` tag and entails no immediate change in the docs. However it could be cumbersome for programmers because they would have to add the `@node` tag for each node annotation.  Wrt the previous example
```Java
/**The type for GetDateTime requests **/
/**@node the date in milliseconds */
type GetDateTimeRequest:long { 
    .format?:string //< .format is the patterns string of the response
                    //<@node the date format in date and time pattern string
}
```
The rendering would be the same as above.
</li>
<li> a third option (that I quickly discarded because it seems too specific to me) is to use an `@root` tag. The tag works like the `@node` tag but is can be used only to attribute to the root the comments around it. 
```Java
/**The type for GetDateTime requests **/
/**@root the date in milliseconds */
type GetDateTimeRequest:long { 
    .format?:string //< the date format in date and time pattern string
}
```
The solution is minimal but to me it seems too specific (but I am open to discussion). In this last case we would not support general comments spread on nodes (which is not necessarily a feature we desperately want :) )
</li>

mawolf:
All the suggested solutions seem fine to me... so please let me know if I should implement all of them, or only on! :-)

klag:
Sorry but I don't like the idea to introduce another annotation.
For me the most simple solution is:
- the comment above the type declaration is about the all type in general-
- the comment on the right of the node is about the node (also for the root).
So at the end should be:
/**The type for GetDateTime requests **/
type GetDateTimeRequest:long {  // the date in milliseconds 
    .format?:string                           //the date format in date and time pattern string
}

fmontesi:
Hi all,

I would expect to be able to write comments as in @klag 's example. I appreciate that @thesave 's suggestions avoid changing the parser and the AST, but it looks like I would need to read up on how our documenter works to understand how to use it. I would prefer it to be so simple that one doesn't need to read anything and just get started by seeing an example.

However, I don't see how to achieve this without changing the AST. I can see two solutions:
1 - The "clean" solution would be to make the native type of a `TypeInlineDefinition` an OLSyntaxNode. This would, however, cause a lot of changes in Jolie and tools. :-(
2 - The "dirty" solution would be to add a field `private String nativeTypeDocumentation;` and the related methods `public String nativeTypeDocumentation()` and `public void setNativeTypeDocumentation( String documentation )` to `TypeInlineDefinition`. This way, the parser could insert any comment that happens after the native type has been parsed but *before* any subnode has been declared to this setter method. Then, joliedoc can extract this special documentation from `TypeInlineDefinition` whenever it meets one.

I think that option 2 makes more sense now. Comments?

mawolf:
Sorry for the delay guys! Have been quite busy at work. :-/ But will try to have this done in this coming week.

Just to be sure: are we completely skipping `//<`?

fmontesi:
We've discussed it over Skype, and we would like to have  //<  :-)

Is my option 2 understandable?

fmontesi:
Hi all,

Any progress on this? It looked almost done! :-)

- Fabrizio

fmontesi:
Dear all,

How's this going? It looked almost done!

Cheers,
Fabrizio

mwallnoefer:
Any news here otherwise I propose to close it (a year of inactivity)?

fmontesi:
This is now supported. :-)

PR #99, opened 2016-10-10
Scheduling and Cancelling timeouts

Tiggles:
Added ability to Schedule a Timeout as well as cancelling it using a UUID.

Also minor logic revision in predicate for language selection.

### DIFF
--- a/include/time.iol
+++ b/include/time.iol
@@ -93,9 +93,19 @@ type DateTimeType:void{
 	.second:int
 }
 
+
 type SetNextTimeOutRequest: int {
 	.operation?: string
 	.message?: undefined
+}
+
+type ScheduleTimeOutRequest: int {
+	.operation?: string
+	.message?: undefined
+	/*
+	*	Possible values are DAYS, HOURS, MICROSECONDS, MILLISECONDS, MINUTES, NANOSECONDS, SECONDS 
+	*/
+	.timeunit?: string
 }
 
 interface TimeInterface{
@@ -106,7 +116,11 @@ interface TimeInterface{
 		  ( default: timeout )
 		*/
 		setNextTimeout(SetNextTimeOutRequest), 
-		setNextTimeoutByDateTime, setNextTimeoutByTime
+		setNextTimeoutByDateTime, setNextTimeoutByTime,
+		/**!
+		* Cancels a time from a UUID created from #scheduleTimeout
+		*/
+		cancelTimeout(string),
 	RequestResponse:
 		getCurrentDateTime(CurrentDateTimeRequestType)(string), sleep,
 
@@ -136,7 +150,11 @@ interface TimeInterface{
 		getTimeDiff(GetTimeDiffRequest)(int),
 		getTimeFromMilliSeconds(int)(TimeValuesType),
 		getTimestampFromString(GetTimestampFromStringRequest)(long) throws FaultException,
-		getDateTimeValues(GetTimestampFromStringRequest)(DateTimeType) throws FaultException
+		getDateTimeValues(GetTimestampFromStringRequest)(DateTimeType) throws FaultException,
+		/**!
+		* Schedules a timeout, which can be cancelled using #cancelTimeout from the returned string. Default .timeunit value is SECONDS.
+		*/
+		scheduleTimeout(SetNextTimeOutRequest)(string),
 }
 
 outputPort Time {
--- a/include/time.iol
+++ b/include/time.iol
@@ -103,7 +103,7 @@ type ScheduleTimeOutRequest: int {
 	.operation?: string
 	.message?: undefined
 	/*
-	*	Possible values are DAYS, HOURS, MICROSECONDS, MILLISECONDS, MINUTES, NANOSECONDS, SECONDS 
+	*	Possible values are DAYS, HOURS, MICROSECONDS, MILLISECONDS, MINUTES, NANOSECONDS, SECONDS. If the value is not set or recognized, it will default to MILLISECONDS
 	*/
 	.timeunit?: string
 }
@@ -120,7 +120,7 @@ interface TimeInterface{
 		/**!
 		* Cancels a time from a UUID created from #scheduleTimeout
 		*/
-		cancelTimeout(string),
+		cancelTimeout(long),
 	RequestResponse:
 		getCurrentDateTime(CurrentDateTimeRequestType)(string), sleep,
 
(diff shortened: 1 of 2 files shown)


include/time.iol
```diff
@@ -93,11 +93,21 @@ type DateTimeType:void{
...
 	.message?: undefined
 }
 
+type ScheduleTimeOutRequest: int {
+	.operation?: string
+	.message?: undefined
+	/*
+	*	Possible values are DAYS, HOURS, MICROSECONDS, MILLISECONDS, MINUTES, NANOSECONDS, SECONDS 
+	*/
+	.timeunit?: string
```
fmontesi: Is MILLISECONDS the default, since the field is optional?

fmontesi: Ah, I see now that it's in seconds, from the doc below. Please make it explicit here.

Isn't MILLISECONDS usually the default in this stuff?

javaServices/coreJavaServices/src/joliex/util/TimeService.java
```diff
@@ -415,4 +422,38 @@ public Long getTimestampFromString( Value request )
 			throw new FaultException( "InvalidTimestamp", pe );
 		}
 	}
+
+	@RequestResponse
+	public String scheduleTimeout( Value request ) {
```
fmontesi: Indentation is inconsistent with the rest of the code (I think I can send you a conf package if you're using Netbeans).

javaServices/coreJavaServices/src/joliex/util/TimeService.java
```diff
@@ -289,7 +296,7 @@ public Value getDateTimeValues( Value request )
 			}
 			if ( request.hasChildren( "language" ) ) {
 				String language = request.getFirstChild( "language" ).strValue();
-				if ( language.equals( "us" ) ) {
+				if ( language.equals( "uk" ) ) {
```
fmontesi: Huh? Why is this in the same commit?

javaServices/coreJavaServices/src/joliex/util/TimeService.java
```diff
@@ -415,4 +422,38 @@ public Long getTimestampFromString( Value request )
 			throw new FaultException( "InvalidTimestamp", pe );
 		}
 	}
+
+	@RequestResponse
+	public String scheduleTimeout( Value request ) {
+		final String uuid = UUID.randomUUID().toString();
```
fmontesi: UUID generation can be slow (requires random number generation). Wouldn't an AtomicLong suffice here instead of using UUIDs?

javaServices/coreJavaServices/src/joliex/util/TimeService.java
```diff
@@ -415,4 +422,38 @@ public Long getTimestampFromString( Value request )
...
+			if ( tasks.contains( uuid ) ) {
+				sendMessage(CommMessage.createRequest(request.getFirstChild("operation").strValue(), "/", request.getFirstChild("message")));
+				tasks.remove(uuid);
+			}
+		}, request.intValue(), unit );
+		return uuid;
+	}
+
+	public void cancelTimeout( Value request ) {
+		tasks.remove( request.strValue() );
```
fmontesi: But the task is still running in the executor, which could prevent the Jolie program from terminating in case we're done with the service. `executor.schedule` actually returns a `ScheduleFuture` that supports a `cancel` method that you can use to cancel the task.

fmontesi:
Thanks! It kinda looks like we could get rid of our timer thread and reuse your executor by having a special identifier for the timeout of setNextTimeout (e.g., the long value 0L).

@klag I remember you were looking at `time.iol`. Do these new operations match what you needed?

javaServices/coreJavaServices/src/joliex/util/TimeService.java
Tiggles: Is this in regards to the curly brace, or am I missing something?

fmontesi:
@klag any news?

klag:
Very nice! I like it!

javaServices/coreJavaServices/src/joliex/util/TimeService.java
```diff
@@ -415,4 +423,44 @@ public Long getTimestampFromString( Value request )
...
+	@RequestResponse
+	public Long scheduleTimeout( Value request )
+	{
+		final long timeoutId = atomicLong.getAndIncrement();
+		TimeUnit unit;
+
+		if (request.hasChildren( "timeunit" )) {
+			try {
+				unit = TimeUnit.valueOf( request.getFirstChild( "timeunit" ).strValue().toUpperCase() );
+			} catch ( Exception e ) {
```
fmontesi: This should probably throw an exception to the caller rather than succeeding silently.

javaServices/coreJavaServices/src/joliex/util/TimeService.java
```diff
@@ -415,4 +423,44 @@ public Long getTimestampFromString( Value request )
...
+				System.err.println( e );
+				System.err.println( "Defaulting to MILLISECONDS" );
+				unit = TimeUnit.MILLISECONDS;
+			}
+		} else {
+			unit = TimeUnit.MILLISECONDS;
+		}
+
+		ScheduledFuture scheduledFuture = executor.schedule( () -> {
+				sendMessage( CommMessage.createRequest(request.getFirstChild( "operation" ).strValue(), "/", request.getFirstChild( "message" )));
```
fmontesi: if the operation is not specified, it should be "timeout"

include/time.iol
```diff
@@ -106,7 +116,11 @@ interface TimeInterface{
 		  ( default: timeout )
 		*/
 		setNextTimeout(SetNextTimeOutRequest), 
-		setNextTimeoutByDateTime, setNextTimeoutByTime
+		setNextTimeoutByDateTime, setNextTimeoutByTime,
+		/**!
+		* Cancels a timeout from a UUID created from #scheduleTimeout
```
fmontesi: Not a UUID anymore.

javaServices/coreJavaServices/src/joliex/util/TimeService.java
```diff
@@ -415,4 +423,44 @@ public Long getTimestampFromString( Value request )
...
+	public void cancelTimeout( Value request )
+	{
+		long timeoutId = request.longValue();
+		synchronized (lock) {
+			ScheduledFuture scheduledFuture = scheduledFutureHashMap.get( timeoutId );
+			if ( scheduledFuture != null ) {
+				scheduledFutureHashMap.get(timeoutId).cancel(false);
+				scheduledFutureHashMap.remove(timeoutId);
+			}
+			else {
```
fmontesi: This makes the operation fail silently without the client knowing.

I think that it makes sense not to throw an exception in case the timer has already been cancelled, since execution of timers is concurrent and the client cannot really know if the timer expired already or not.

I would return a boolean value, though: `true` if the timer has been cancelled, `false` otherwise.

@jolie/developers comments?

javaServices/coreJavaServices/src/joliex/util/TimeService.java
```diff
@@ -415,4 +423,44 @@ public Long getTimestampFromString( Value request )
...
+				sendMessage( CommMessage.createRequest(request.getFirstChild( "operation" ).strValue(), "/", request.getFirstChild( "message" )));
+		}, request.intValue(), unit );
+		scheduledFutureHashMap.put(timeoutId, scheduledFuture);
+		return timeoutId;
+	}
+
+	public void cancelTimeout( Value request )
+	{
+		long timeoutId = request.longValue();
+		synchronized (lock) {
```
fmontesi: You don't need a lock.

You can just

```
boolean retVal;
ScheduledFuture f = scheduledFutureHashMap.remove( timeoutId );
if ( f != null ) {
    retVal = f.cancel( false );
} else {
    retVal = false;
}
```

PR #105, opened 2017-01-31
postponed encoded content creation in order to allow field removing wh…

abjunior92:
…en alias are parsed

### DIFF
--- a/extensions/http/src/jolie/net/HttpProtocol.java
+++ b/extensions/http/src/jolie/net/HttpProtocol.java
@@ -871,7 +871,7 @@ public class HttpProtocol extends CommProtocol implements HttpUtils.HttpProtocol
 		Method method = send_getRequestMethod( message );
 		String charset = HttpUtils.getCharset( getStringParameter( Parameters.CHARSET, "utf-8" ), null );
 		String format = send_getFormat();
-		EncodedContent encodedContent = send_encodeContent( message, method, charset, format );
+		String contentType = null;
 		StringBuilder headerBuilder = new StringBuilder();
 
 		if ( inInputPort ) {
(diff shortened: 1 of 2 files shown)


extensions/http/src/jolie/net/HttpProtocol.java
```diff
@@ -1339,8 +1340,20 @@ public CommMessage recv_internal( InputStream istream, OutputStream ostream )
...
+                                responseHeader = getOperationSpecificStringParameter(inputId, Parameters.RESPONSE_HEADER);
+                            }
+                            else {
+                                responseHeader = getStringParameter(Parameters.RESPONSE_HEADER);
+                            }
+                            for(Entry<String, String> param : message.properties()){
+                                decodedMessage.value.getFirstChild(responseHeader).getFirstChild(param.getKey()).setValue(param.getValue());
+                            }
+                        }
+                        decodedMessage.value.getFirstChild(responseHeader).getFirstChild("statusCode").setValue(message.statusCode());
```
fmontesi: Indentation seems wrong.

extensions/http/src/jolie/net/HttpProtocol.java
```diff
@@ -191,6 +191,7 @@
 		private static final String COMPRESSION_TYPES = "compressionTypes";
 		private static final String REQUEST_COMPRESSION = "requestCompression";
 		private static final String FORMAT = "format";
+                private static final String RESPONSE_HEADER = "response_header";
```
fmontesi: Indentation seems wrong.

Also, response_header should be plural (since we have more than one header) and it looks like we're using camelCase in the rest of the parameters, so that gives us "responseHeaders". Can you switch?

fmontesi:
Mmh, I think I still see spaces instead of tabs in HttpProtocol.java . :-)

abjunior92:
Maybe because i’m writing over Atom and not Netbeans?

fmontesi:
Any news on this?

abjunior92:
Indentetion is still wrong? I believe it was fixed after my last commit…or not?

PR #106, opened 2017-03-16
Flags added to jolie2surface

Tiggles:
Added the three flags

-noOutputPort
-noLocation
-noProtocol

allowing for omitting the outputport or location/protocol from jolie2surface.


fmontesi:
Looks good to me. @klag , does it break anything in your use cases?

klag:
it should be ok.

2017-03-20 15:14 GMT+01:00 Fabrizio Montesi <notifications@github.com>:

PR #110, opened 2017-03-28
Added fabs operation for Doubles in the Math service

Tiggles:
The Math service didn't have a corresponding operation for getting the absolute value for Doubles, "fabs" might just not be an ideal name as Jolie only seems to support doubles.


mwallnoefer:
I guess that this pull request should just contain commit 861bd81.
Please create separate pull requests for the other patches, since in this form we cannot review.

PR #115, opened 2017-06-08
1.5.1

klag:
- Bug fixes for allowing importing of jolie interpreter within a jboss application server. 
- added hook operator (=$) for allowing dynamic data structure navigation
- added plain_xml_manager.iol javaservice for managing plain XML documents without loosing any information


fmontesi:
Hi @klag 

I finally got around writing some lines about pull requests. Could you refactor the pull request based on https://jolielang.gitbooks.io/developer-guide/content/how-to-make-a-pull-request.html

Cheers,
Fabrizio

klag:
here there are: 
    - Bug fixes for allowing importing of jolie interpreter within a jboss application server.
    - added hook operator (=$) for allowing dynamic data structure navigation
    - added plain_xml_manager.iol javaservice for managing plain XML documents without loosing any information
We need to decide what we want to include and what we do not want. I think the first item is quite important to be included. We discussed the second one some moths ago but we did not decide anything about it. Finally the third item is strictly related to the second one because it allows for a full navigation of a xml document by exploring all its details and without exploiting a service recursion call. As I noticed in the discussion we had, without a specific operator like the hook one, so far it is not possible to recursively navigate a tree structure within a single session but we always need to exploit an external recursion call to a service which always imply a copy of the structure in a new message.

mwallnoefer:
Okay, n° 1 seems okay.
Regarding points n°2 and 3 it depends if there exist production Jolie instances which still need to run under Java 6 or 7. If the answer is yes then I would merge them, otherwise probably not.

mwallnoefer:
So what to do here? Either merge it into the java-6 branch otherwise I propose to close it.

klag:
I think we can close the conversation wothout merging anything. But plesse do not erase the branch. There is the implementation of the hook operator, I would like to keep it until we have functions in jolie

PR #116, opened 2017-06-19
Added check for localsocket in http calls

klag:


extensions/soap/src/jolie/net/SoapProtocol.java
```diff
@@ -786,6 +827,39 @@ public void send_internal( OutputStream ostream, CommMessage message, InputStrea
 					}
 					SOAPElement opBody = soapBody;
 					if ( wrapped ) {
+						List<ExtensibilityElement> listExt;
+						if ( received ) {
+							listExt = getWSDLPort().getBinding().getBindingOperation( message.operationName(), null, null).getBindingOutput().getExtensibilityElements();
+						} else {
+							listExt = getWSDLPort().getBinding().getBindingOperation( message.operationName(), null, null).getBindingInput().getExtensibilityElements();
+						}
+						;
```
fmontesi: empty statement? Remove if so.

extensions/http/src/jolie/net/HttpProtocol.java
```diff
@@ -878,7 +883,7 @@ public void send_internal( OutputStream ostream, CommMessage message, InputStrea
 		if ( inInputPort ) {
 			// We're responding to a request
 			send_appendResponseHeaders( message, headerBuilder );
-                        send_appendResponseUserHeader(message, headerBuilder);
+			send_appendResponseUserHeader(message, headerBuilder);
```
fmontesi: indentation (spacing).

extensions/http/src/jolie/net/HttpProtocol.java
```diff
@@ -888,7 +893,7 @@ public void send_internal( OutputStream ostream, CommMessage message, InputStrea
 					contentType = ContentTypes.APPLICATION_JSON;
 				}
 			}
-                        send_appendRequestUserHeader(message, headerBuilder);
+			send_appendRequestUserHeader(message, headerBuilder);
```
fmontesi: indentation.

extensions/http/src/jolie/net/HttpProtocol.java
```diff
@@ -936,7 +941,11 @@ private void parseXML( HttpMessage message, Value value, String charset )
 	private static void parseJson( HttpMessage message, Value value, boolean strictEncoding, String charset )
 		throws IOException
 	{
-		JsUtils.parseJsonIntoValue( new InputStreamReader( new ByteArrayInputStream( message.content() ), charset ), value, strictEncoding );
+		try {
+			JsUtils.parseJsonIntoValue( new InputStreamReader( new ByteArrayInputStream( message.content() ), charset ), value, strictEncoding );
```
fmontesi: Shouldn't this throw a fault? This way we get a string message and maybe the programmer was not expecting it in the type, hence raising an unexpected TypeMismatch fault.

extensions/http/src/jolie/net/HttpProtocol.java
```diff
@@ -202,6 +203,7 @@
 		private static final String CONTENT_DISPOSITION = "contentDisposition";
 		private static final String DROP_URI_PATH = "dropURIPath";
 		private static final String CACHE_CONTROL = "cacheControl";
+		private static final String STRING_RAW_CONTENT = "stringRawContent";
```
fmontesi: I'm having a really hard time finding a good name for this parameter.. :-)

Raw is a bit confusing because we have the raw data type.

What about `doNotDecode` or `getStringContent`? They are all bad ... ;-) @jolie/developers ideas for a better name for this parameter?

mwallnoefer: I do not like it so much, this new parameter which switches off the correct content parsing. Is it not possible to put these messages under the content type **application/octet-stream**?

If not I would call such a parameter _forceRaw_.

fmontesi: Me neither, and indeed this is just a solution for when we have to deal with buggy software, basically. In this case, @klag is dealing with a software that sends a "stream" of separate JSON objects through a single HTTP response (a bit of a hack). So the parsing needs to be handled manually. The application is not using application/octet-stream as content type, and instead triggers our JSON parser, which gets angry because it expects a single JSON message.

To my knowledge, sending a stream of JSON objects with application/json is just plain wrong: http://www.ietf.org/rfc/rfc4627.txt

But maybe I'm misreading the specs?

`forceRaw` does not sound too bad actually. Could `forceString` be better maybe? We're still doing the charset decoding, so it's not really "raw". (And we're not returning something with the `raw` data type.)

Could it ever make sense to return the content as a `raw` (byte array)? In that case, we may want a parameter that can take `string` and `raw` as values and then act accordingly.

klag: What about `forceContentEnconding` and then specifying `string` or `raw` in asubnode filed called `encodingType` as Fabrizio proposed?

mwallnoefer: Okay, so it works like this (latest RFC 7159: https://tools.ietf.org/html/rfc7159):

> 9.  Parsers
>   A JSON parser transforms a JSON text into another representation.  A
>   JSON parser MUST accept all texts that conform to the JSON grammar.
>   A JSON parser MAY accept non-JSON forms or extensions.

> 2.  JSON Grammar
>   A JSON text is a sequence of tokens.  The set of tokens includes six
>   structural characters, strings, numbers, and three literal names.
>   A JSON text is a serialized value.  ...

> 3.  Values
>   A JSON value MUST be an object, array, number, or string, or one of
>   the following three literal names:
>   false null true

So Fabrizio you are right, there cannot be transferred **more than one array, object, value** per HTTP request. But a parser may be open for extensions, unlike the generator (_"The resulting text MUST strictly conform to the JSON grammar"_).

fmontesi: Thank you both.

Mmh, then I'd go for something like `.forceContentDecoding:string` (since we're decoding, not encoding) that can be set to either "string" or "raw". That way, it is similar to parameter `format`. Unless there are reasons for thinking that this is different, in that case we could use a sub-subnode as @klag said to specify extra parameters?

klag: No, i think `.forceContentDecoding: string` is enough. If you agree I proceed with the change

fmontesi: Sounds good then, go ahead. @jolie/developers please pitch in if you have better ideas.

klag: done

PR #132, opened 2018-06-08
Soap fix

thesave:
Integrated soap local service into ant build and test tasks


build.xml
```diff
@@ -360,24 +362,28 @@ Please make sure that:
...
-    	</exec>
-    </target>
+			<exec executable="sh" dir="test">
+				<env key="JOLIE_HOME" file="dist/jolie" />
+				<arg value="../dist/launchers/unix/jolie" />
+				<arg value="-l" />
+				<arg value="extensions/private/WS-test/dist/WS-test.jar" />
+				<arg value="test.ol" />
+			</exec>
+		</target>
```
fmontesi: This will break the ability of running simply `jolie test.ol`. Please put the library argument you need in the Jolie code (e.g., by using loadLibrary@Runtime or by embedding something and passing the `-l` option in the embedding string).

build.xml
```diff
@@ -360,24 +362,28 @@ Please make sure that:
...
-    </target>
+			<exec executable="cmd" dir="test">
+				<env key="JOLIE_HOME" file="dist\jolie" />
+				<arg value="/c" />
+				<arg value='..\dist\launchers\windows\jolie.bat --charset "utf-8"' />
+				<arg value="/l" />
+				<arg value="extensions\private\WS-test\dist\WS-test.jar" />
+				<arg value="test.ol" />
+			</exec>
+		</target>
```
fmontesi: See the above comment.

build.xml
thesave: Fixed.

PR #148, opened 2018-10-10
added fileTree operation to File Javaservice

klag:


mwallnoefer:
I think this little patch looks okay so I merge it.

fmontesi:
Mmh, looks like this is a bit redundant with `list@File`. Why don't we just pass a boolean parameter to `list@File` that tells it whether it should go recursively inside of subdirectories, like in Unix tools?
There's a lot of shared functionality that would be useful there. For example, one might want the returned paths to match a regex, or get info like size. All these functionalities are provided in `list@Files` and we would have to copy them for `fileTree` in the future.

fmontesi:
Concretely, we could add the parameter `.recursive?:bool` to `list@Files`, which defaults to `false` (such that we do not break old behaviour). What do you think, @mwallnoefer @klag ? Can any of you two do that, if you agree?

mwallnoefer:
Should we revert and let @klag develop a new patch?

klag:
I checked the code of `list` and it returns a vector of nodes `result` where each node contains the name of the file (or directory) and some optional sub-fields for the info. If we introduce the possibility to call it recursively, each node `result` will contain other nodes `result` for the second level of files, the second level of nodes `result` will contain other nodes `result` for the third level of files, and so on. This would be a different response w.r.t. `fileTree` because `fileTree` just returns a list of paths (a vector of strings).
If we want to change such a kind of response structure we need to re-think the response format of the method `list` in general.

Even if it could be of interest to add recursion in the `list` method, I think that `fileTree` offers a different functionality w.r.t. that we would obtain with a recursive `list`.

mwallnoefer:
Nice cleanup, @fmontesi. 
@klag So does Fabrizio's rework address your needs?
So the only thing left to do is to add a recursive traversal to the unit test?

mwallnoefer:
So I tested the _list()_ call (files.ol test) on both Linux and Windows and it seems to work.

fmontesi:
Awesome, thanks! :) 

I'll consider this fixed then for now. @klag feel free to add comments in the future, but if I don't hear anything about this, I'll assume it's ok to include the list@File modification in the next release.

@mwallnoefer could you push the new tests?

klag:
It's ok for me!

mwallnoefer:
The same tests passed on both Linux and Windows, so this should be fine.

But we need to fix some other path problem on Windows in respect to the SOAP tests, which I will post in another thread.

PR #151, opened 2018-10-12
Removed extra & in HTTP query

bmaschio:


mwallnoefer:
I also think that here the risk to break something is higher than the benefit. Okay, aesthetically this trailing ampercent is not very appealing, but technically it is not wrong either I guess?

bmaschio:
@fmontesi @mwallnoefer  

> What if one of the ValueVectors has more than one elements?

When would that happen?
Thanks 
B

fmontesi:
```
request.x[0] = 1;
request.x[1] = 2
```

Then you send `request`.

bmaschio:
and that would be translated in `http://someurl?x=1&x=2` my question is this correct I mean logically?
maybe is a silly question @lsafina @mwallnoefer

mwallnoefer:
As far as I see yes, this is correct (and `http://someurl?x=1&x=2` is correct either)

fmontesi:
Would it?

You use `value.children().entrySet().size()`. Isn't that going to be `1`? I suspect it's going to yield: `http://someurl?x=1`.

You should take the sum of all sizes of all vectors, not just the size of `children()` (which is just the number of vectors inside of `request`).

Maybe I'm misreading the code snippet?

fmontesi:
Hi @bmaschio @mwallnoefer ,

I've made a fix using Iterators: https://github.com/jolie/jolie/commit/25eea6273f71b0e48aa43cf16f4ede57497c35b2

Tests work, but an extra two pairs of eyeballs wouldn't hurt.. please have a look and reopen this if something smells. :-)

I'll wait for your comments before releasing it.

F

mwallnoefer:
Seems correct!

PR #152, opened 2018-10-15
Jolie integration with IoT-related protocols [OFFICIAL]

saltgz:
Stuff added (should be relatively safe to merge):
PubSubCommProtocol
PubSubCommProtocolProtocolFactory
CoapProtocol (and related)
MqttProtocol (and related)
Datagram (and related)

Please be careful with:
NioSocketListener
NioSocketCommChannel
CommChannelFactory
CommCore

Some Fixes:
AutoProtocol
xsom
build.xml
build.config


thesave:
Thanks!

mwallnoefer:
And I hope that also some unit testing covering the new code gets added in order to guarantee the functionality.

thesave:
Hi @mwallnoefer, thanks for taking a look at the code!
Right now this branch is still in development and pretty shaky but, as you mentioned, we will dutifully unit-test the new (and old) functionalities before merging into the master :)

saltgz:
Hi @mwallnoefer, as @thesave mentioned, this is still an ongoing wok on integrating netty technologies into the master branch (**next** 👍 ), and thus testing is our main purpose right now. 
By the way, I don't get your question on compression.. the mods simply states that the compression request is disabled by default.
Comments are more than welcome, and testing help too ;-)

mwallnoefer:
> By the way, I don't get your question on compression.. the mods simply states that the compression request is disabled by default.

And what's the particular reason to change this? I mean HTTP compression was designed to be optional, I mean also all web browsers add the compression request header per default - so it is up to the server to respond compressed or not. 
So did you encounter some difficulties with our implementation?

saltgz:
> So did you encounter some difficulties with our implementation?

[ant_test_log_http_compression_true.txt](https://github.com/jolie/jolie/files/2526847/ant_test_log_http_compression_true.txt)

From the **io.netty.handler.codec.compression.DecompressionException** raised with message _Input is not in the GZIP format_, it seems that the client cannot compress properly. The compression to false fixes the bug, further investigations have to be made.

Again, 
> Comments are more than welcome, and testing help too 😄

mwallnoefer:
I have been the original contributor of the compression feature and never ran into problems with its implementation. It was even enabled on Jolie's homepage back when it was running under Jolie/Leonardo (now it's driven by nginx?).

So it would be nice to discover the incompatibility against netty (which I unfortunately don't know), otherwise we will probably disable it and then it gets never enabled again.

saltgz:
"[Netty in action](https://www.manning.com/books/netty-in-action)" is a good starting point to get to know netty.
Nevertheless, we can always set this as an issue and resolve it later.

PR #153, opened 2018-10-19
added raw support to forceContentDecoding

bmaschio:
As discussed in the mail I added 

    .osc.leggiFatturaFile.forceContentDecoding = "raw"


fmontesi:
Super, thanks! :+1:

PR #156, opened 2018-10-24
Ndjson

bmaschio:


bmaschio:
@fmontesi  not sure what is problem

lib/jolie-js/src/jolie/js/JsUtils.java
```diff
@@ -121,7 +125,51 @@ public static void valueToJsonString( Value value, boolean extendedRoot, Type ty
 			builder.append( '}' );
 		}
 	}
+	public static void valueToNdJsonString( Value value, boolean extendedRoot, Type type, StringBuilder builder ) throws IOException
+	{
+		Map<String, ValueVector> children = value.children();
+		for( Map.Entry<String, ValueVector> child : children.entrySet() ) {
+			for( Value v : child.getValue() ) {
+				if ( v.hasChildren( JSONARRAY_KEY ) ) {
```
fmontesi: Why don't you just call `valueToJsonString` for each item here?

lib/jolie-js/src/jolie/js/JsUtils.java
```diff
@@ -224,4 +272,32 @@ public static void parseJsonIntoValue( Reader reader, Value value, boolean stric
 			throw new IOException( e );
 		}
 	}
+	
+	public static void parseNdJsonIntoValue( Reader reader, Value value, boolean strictEncoding )
+		throws IOException
+	{
+		String stringBuffer = new BufferedReader( (InputStreamReader) reader ).lines().collect( Collectors.joining( "\n" ) );
```
fmontesi: I don't like this explicit cast to `InputStreamReader`. If you need a BufferedReader, get a BufferedReader already in the type of the method.

lib/jolie-js/src/jolie/js/JsUtils.java
```diff
@@ -224,4 +272,32 @@ public static void parseJsonIntoValue( Reader reader, Value value, boolean stric
...
+		String[] stringItemVector = stringBuffer.split( "\n" );
```
fmontesi: So first you collect all the lines in a stringBuffer, and then you split it again? Why don't you just use the stream given by `lines()` in buffered reader and its `forEach` method?

lib/jolie-js/src/jolie/js/JsUtils.java
bmaschio: because is a public method I didn't think you would like it

PR #158, opened 2018-11-02
Ndjson

bmaschio:
ndJson Support


mwallnoefer:
Also here it would have been wise to add at least a basic unit test, since this seems pretty new functionality. Otherwise how we make sure to not break it?

PR #159, opened 2018-11-06
email with attachment

bmaschio:
@fmontesi 
Hey there
I have implemented the capability of sending attachments with the email message, the implementation requires the use of  mailcap file placed in the directory META-INF containg

### image/jpeg;;		x-java-content-handler=com.sun.mail.handlers.image_jpeg

Alternatively, we can add the following code 

```
MailcapCommandMap mc = (MailcapCommandMap) CommandMap.getDefaultCommandMap(); 
mc.addMailcap("text/html;; x-java-content-handler=com.sun.mail.handlers.text_html"); 
mc.addMailcap("text/xml;; x-java-content-handler=com.sun.mail.handlers.text_xml"); 
mc.addMailcap("text/plain;; x-java-content-handler=com.sun.mail.handlers.text_plain"); 
mc.addMailcap("multipart/*;; x-java-content-handler=com.sun.mail.handlers.multipart_mixed"); 
mc.addMailcap("message/rfc822;; x-java-content- handler=com.sun.mail.handlers.message_rfc822"); 
```
What do you think


fmontesi:
I like the defaults more than the mailcap file. We should give the option to set these mailcap things with a command maybe, but let's do that as a separate patch later.

@jolie/developers I'd like some more eyes on this. In particular, at the very least, I'd like opinions on:
- This patch will make all e-mails sent a multipart e-mail. Is that relevant for any use case? At the very least @klag should probably test this patch in a system that uses e-mail.
- Will this thing work with Java 11? I see references to internal sun classes in these mailcap parameters. @thesave Can you try? Do we need to import any libraries from https://javaee.github.io/javamail/ maybe?

mwallnoefer:
I would keep both behaviours: normal and multi-part. If there is just one raw content, I suggest to keep it as 'normal'/raw to not raise the complexity unnecessarily.

bmaschio:
@jolie/developers So how shell I proceed ?

thesave:
@fmontesi I imported the changes in SMTPService.java from the mail branch and `ant`-compiled the whole project under Java 11.
Everything went well, no errors or warnings on deprecated libraries over the `javax.mail.*` APIs.

bmaschio:
Wrong closing request

PR #174, opened 2019-04-12
Removed private folder for project joliedummy

klag:


fmontesi:
This looks like a very old pull request for just removing a few private files. Can you open a separate one?

PR #208, opened 2019-11-08
Rest correct

bmaschio:
Only jolier commits


bmaschio:
impact only on jolier

PR #214, opened 2020-01-05
Tracer

klag:
First version of jolietraceviewer done


extensions/http/src/main/java/jolie/net/HttpProtocol.java
```diff
@@ -856,6 +851,10 @@ private void send_appendGenericHeaders(
 				compression = false;
 			}
 			if ( compression ) {
+				if ( !(Interpreter.getInstance().tracer() instanceof DummyTracer) ) {
```
fmontesi: Never do this check, it breaks abstraction. The API of the tracer is designed to be efficient and not require these checks at the same time. That's why we use lambdas. In this case, for example, you can just do:

```
Interpreter.getInstance().tracer().trace( () -> {
    final String traceMessage = encodedContent.content.toString(charset);
    return new ProtocolTraceAction(ProtocolTraceAction.Type.HTTP, "HTTP COMPRESSING MESSAGE", message.resourcePath(), traceMessage, null );
} );
```

extensions/http/src/main/java/jolie/net/HttpProtocol.java
```diff
@@ -918,6 +924,12 @@ public void send_internal( OutputStream ostream, CommMessage message, InputStrea
 		headerBuilder.append( HttpUtils.CRLF );
 
 		send_logDebugInfo( headerBuilder, encodedContent, charset );
+
+		if ( !(Interpreter.getInstance().tracer() instanceof DummyTracer) ) {
```
fmontesi: See previous comment on this check.

extensions/http/src/main/java/jolie/net/HttpProtocol.java
```diff
@@ -1331,7 +1341,19 @@ public CommMessage recv_internal( InputStream istream, OutputStream ostream )
...
+			boolean showContent = false;
+			if (
+					getParameterFirstValue( Parameters.DEBUG ).getFirstChild( "showContent" ).intValue() > 0
+							&& message.size() > 0
+			) { showContent = true; }
+			Interpreter.getInstance().logInfo( getDebugMessage( message, charset, showContent ));
+		}
+
+		// tracer
+		if ( !(Interpreter.getInstance().tracer() instanceof DummyTracer) ) {
```
fmontesi: Ditto above.

klag: done

PR #227, opened 2020-05-08
raise a warning when using a variable named as a scope

kicito:
log deprecate if there is an usage of scope's variable name in LHS of statements.

changes OOITBuilder:
- new overload function buildVariablePath accepts new boolean indicates if building path is LHS expression and perform checks of root varpath name to scope id
- all visit methods for assignment related node uses methods created above
- new class scope field currScopeId to store scopeId context
- warning method in OOITBuilder

### DIFF
--- a/jolie/src/main/java/jolie/OOITBuilder.java
+++ b/jolie/src/main/java/jolie/OOITBuilder.java
@@ -1049,7 +1049,7 @@ public class OOITBuilder implements OLVisitor
 			
 		AssignmentProcess p = 
 			new AssignmentProcess(
-				buildVariablePath( n.variablePath() ),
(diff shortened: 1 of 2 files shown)


fmontesi:
Thank you @kicito ! :-)

I'd rather not have it in the "compiler" (the OOITBuilder), but rather in a code analyser like the SemanticVerifier. Could you move the logic there? When you meet an assignment, a link, or a deepcopy statement, you could just check that the variable path on the left doesn't start with a name of the variable that is also a scope name.

kicito:
Yes, for some reason I didn't think about it :sweat_smile:.

fmontesi:
Well done, thank you very much! :)

PR #231, opened 2020-05-22
name clash checking for variable/nested scope 

kicito:
name clashing checking between variable and scope name support checking for nested scope


fmontesi:
Awesome, thanks. :-)

PR #233, opened 2020-05-26
Number ops hierarchy

### LINKED ISSUE #232: Behaviour of double * int different from int * double
## Reproduce

Consider the following code (actual output in the comments):

```jolie
include "console.iol"

main
{
  a = 0.1 * 1000
  b = 1000 * 0.1
  println@Console( a == b )() # false
  c = double( 1000 ) * 0.1
  println@Console( a == c )() # true
}
```

## Question

Is this **asymmetric behaviour** wanted? 
Not sure this should be a bug, I was just wondering why this choice.

## Context

- MacOSX 10.15.5
- Jolie interpreter 1.9.0
- OpenJDK Runtime Environment (build 13+33)

kicito:
fix #232 

add number type hierarchy checking and conversion to both side of operation.

hierarchy is following:
int < long < double


fmontesi:
Dear @jolie/developers,

This changes the semantics of Jolie's arithmetic operators (for the better, since they would now be commutative for numbers at least). In the future, we should probably do the same for strings.

Can you all have a look? I plan on merging this on Monday, unless there are specific objections.

fmontesi:
@kicito and @klag, we have this test failing:

Jun 15, 2020 12:17:51 PM zip_utils.ol
SEVERE: [zip_utils.ol] Request-Response process for test threw an undeclared fault for that operation (TypeMismatch), throwing TypeMismatch
Jun 15, 2020 12:17:51 PM zip_utils.ol
INFO: [zip_utils.ol] Output message TypeMismatch (stringToRaw@Converter): Invalid native type for node #Message: expected STRING, found java.lang.Integer
Jun 15, 2020 12:17:51 PM test.ol
INFO: [test.ol] Internal server error

Could you check whether it's related to this pull request or not?

fmontesi:
This seems to be the offending line in test/library/zip_utils.ol

```
content = "this is a test " + i;
```

@kicito, does the test run on your machine with your patches?

kicito:
I didn't added the conversion on string type on initial commit. With the recent commit, which support string type checking to both side of add operation. It passes case you've mention now.

PR #239, opened 2020-06-05
Unbundle command line parser

klag:


jolie/src/main/java/jolie/runtime/embedding/JolieServiceLoader.java
```diff
@@ -45,9 +49,26 @@ public JolieServiceLoader( Expression channelDest, Interpreter currInterpreter,
...
+	}
+
+	public JolieServiceLoader( String code, Expression channelDest, Interpreter currInterpreter )
+		throws IOException {
+		super( channelDest );
+		InterpreterParameters interpreterParameters = new InterpreterParameters(
+			currInterpreter.optionArgs(),
+			currInterpreter.getInterpreterParameters().includePaths(),
+			currInterpreter.getInterpreterParameters().libUrls(),
+			new File( "#native_code" ),
```
fmontesi, line 66: You should generate something different for each embedded piece of code, or the debugging messages won't make much sense.

jolie/src/main/java/jolie/CommandLineParser.java
```diff
@@ -94,7 +94,7 @@
 	 * 
 	 * @return the arguments passed to the JOLIE program.
 	 */
-	public final String[] arguments() {
+	private final String[] arguments() {
```
fmontesi, line 97: Do you still need all these private accessor methods? You could just access the fields directly when you create the InterpreterParameters object, since they're private now.

jolie/src/main/java/jolie/CommandLineParser.java
```diff
@@ -94,7 +94,7 @@
 	 * 
 	 * @return the arguments passed to the JOLIE program.
 	 */
```
fmontesi, line 96: Please copy the javadoc comments to the corresponding methods in InterpreterParameters, or we will lose documentation. (I'd like to keep at least the one we have... :-) )

klag:
it should be ok now

PR #242, opened 2020-06-15
String constrained nativetype

klag:
Refactoring of NativeType (added new type TypeNative and renamed NativeType into NativeTypeEnum) in order to manage native type constraints.

Added the possibility to define the string length into a type definition .

`type MyType: string( 4 ) {
     f1: string( 10 )
}`

where 4 and 10 define the maximun length of the string


extensions/jsonrpc/src/main/java/jolie/net/JsonRpcProtocol.java
```diff
@@ -140,8 +141,8 @@ public void send_internal( OutputStream ostream, CommMessage message, InputStrea
 		Type operationType = Type.UNDEFINED;
 		String originalOpName = message.operationName();
 		Map< String, Type > subTypes = new HashMap<>();
-		subTypes.put( "jsonrpc", Type.create( NativeType.STRING, new Range( 1, 1 ), false, null ) );
-		subTypes.put( "id", Type.create( NativeType.INT, new Range( 0, 1 ), false, null ) );
+		subTypes.put( "jsonrpc", Type.create( new BasicType( NativeType.STRING ), new Range( 1, 1 ), false, null ) );
```
fmontesi, line 144: BasicType.of( .. ) like Long.valueOf( ... )

PR #244, opened 2020-06-15
added charset control on connection

bmaschio:
@fmontesi @klag check if everything works...


fmontesi:
You might also want to add useUnicode, but *only* for mysql. (But I suggest this is a separate patch, since this pull request can already go through.)

@klag let me know when you've had a look.

klag:
ok for me

PR #253, opened 2020-06-30
Parametric port

kicito:
This PR changes PortInfo fields for location and protocol to an Expression.

The flow of parsing these fields is following

- OLParser: parses the fields using parseBasicExpression(), the protocolId and protocolConfiguration fields are combined into one single field.
- Optimizer: Optimizes both expressions 
- SemanticVerifier: verifies instance of expression as following
  - location field has to be only an ConstantStringExpression, which its value is a valid URI format.
  - protocol field can be either ConstantStringExpression, VariablePathNode( when parsing protocol identifier ), InlineTreeExpression
- OOITBuilder: evaluate expressions and create Port/CommCoreListener Object accordingly

This PR also introduce an utility class: ModuleSystemUtil which implement an utility function to make Jolie's Module System backward compatible to old syntax.


javaServices/coreJavaServices/src/main/java/joliex/meta/MetaJolie.java
```diff
@@ -571,8 +571,8 @@ private Value getPort( PortInfo portInfo, List< InterfaceDefinition > interfaces
 			} else {
 				response.getFirstChild( "location" ).setValue( "local" );
 			}
-			if( port.protocolId() != null ) {
-				response.getFirstChild( "protocol" ).setValue( port.protocolId() );
+			if( port.protocol() != null ) {
+				response.getFirstChild( "protocol" ).setValue( port.protocol() );
```
fmontesi, line 575: toString() ?

javaServices/coreJavaServices/src/main/java/joliex/meta/MetaJolie.java
```diff
@@ -584,8 +584,8 @@ private Value getPort( PortInfo portInfo, List< InterfaceDefinition > interfaces
...
```
fmontesi, line 588: .toString()

libjolie/src/main/java/jolie/lang/parse/OLParseTreeOptimizer.java
```diff
@@ -157,37 +157,47 @@ public void visit( CorrelationSetInfo p ) {
 
 		@Override
 		public void visit( OutputPortInfo p ) {
-			if( p.protocolConfiguration() != null ) {
-				p.protocolConfiguration().accept( this );
-				p.setProtocolConfiguration( currNode );
+			if( p.protocol() != null ) {
+				p.protocol().accept( this );
```
fmontesi, line 161: You can just call `optimize`

libjolie/src/main/java/jolie/lang/parse/OLParseTreeOptimizer.java
```diff
@@ -157,37 +157,47 @@ public void visit( CorrelationSetInfo p ) {
...
-					iport.setDocumentation( p.getDocumentation() );
-				}
-				iport.operationsMap().putAll( p.operationsMap() );
-				iport.getInterfaceList().addAll( p.getInterfaceList() );
-				programChildren.add( iport );
-			} else {
-				programChildren.add( p );
+			OLSyntaxNode protocol = null, location = null;
+			if( p.protocol() != null ) {
+				p.protocol().accept( this );
```
fmontesi, line 175: you can just call `optimize`

libjolie/src/main/java/jolie/lang/parse/util/ModuleSystemUtil.java
```diff
@@ -0,0 +1,56 @@
...
+		// expression
+		// case some.proc -> do noting. return VariablePathNode
+		// case "http" {} -> do noting. return InlineTreeExpression
+
+		// case 1
+		if( node instanceof VariableExpressionNode ) {
+			VariableExpressionNode varExprNode = (VariableExpressionNode) node;
+			VariablePathNode varPathNode = varExprNode.variablePath();
+			if( varPathNode.path().size() == 1 ) {
+				return varPathNode.path().get( 0 ).key();
```
fmontesi, line 34: What if this was actually intended as a variable path, because it's a parameter of the service?

Maybe check that it's one of the supported protocols?

libjolie/src/main/java/jolie/lang/parse/util/ModuleSystemUtil.java
```diff
@@ -0,0 +1,56 @@
...
+		// case 2 checks node is an InlineTreeExpressionNode with VariableExpressionNode as root expression
+		if( node instanceof InlineTreeExpressionNode && ((InlineTreeExpressionNode) node)
+			.rootExpression() instanceof VariableExpressionNode ) {
+			InlineTreeExpressionNode inlineTreeNodeProtocol = (InlineTreeExpressionNode) node;
+			if( inlineTreeNodeProtocol.rootExpression() instanceof VariableExpressionNode ) {
+				String protocolSymbolStr =
+					((VariableExpressionNode) inlineTreeNodeProtocol.rootExpression())
+						.toString();
+				node = new InlineTreeExpressionNode( inlineTreeNodeProtocol.context(),
+					new ConstantStringExpression( inlineTreeNodeProtocol.context(),
```
fmontesi, line 47: same comment as above. You might want to put the previous if (lines 30-36) in a separate method that you can also call here.

javaServices/coreJavaServices/src/main/java/joliex/meta/MetaJolie.java
```diff
@@ -403,8 +403,8 @@ private Value getOutputPort( OutputPortInfo portInfo ) {
 		} else {
 			response.getFirstChild( "location" ).setValue( "undefined" );
 		}
-		if( portInfo.protocolId() != null ) {
-			response.getFirstChild( "protocol" ).setValue( portInfo.protocolId() );
+		if( portInfo.protocol() != null ) {
+			response.getFirstChild( "protocol" ).setValue( portInfo.protocol().toString() );
```
fmontesi, line 407: I wouldn't invoke OLSyntaxNode.toString(), this is misleading, since I'd expect an InlineTreeExpression to print out the entire tree, not just the root.

Consider keeping the `protocolId` method to do this maybe?

libjolie/src/main/java/jolie/lang/parse/ast/InputPortInfo.java
```diff
@@ -84,4 +87,23 @@ public OLSyntaxNode location() {
...
+			p = ((ProductExpressionNode) prodVal).operands().iterator().next().value();
+		}
+
+		if( p instanceof InlineTreeExpressionNode ) {
+			return ((InlineTreeExpressionNode) p)
+				.rootExpression();
+		} else {
+			return p;
+		}
+	}
```
kicito, line 108: I am not sure that, in order to bring back protocolId() in this commit, I should modify OLParser to separately parsing protocolId and protocol configuration as we had before this commit? Is this method make sense for you?

jolie/src/main/java/jolie/OOITBuilder.java
```diff
@@ -565,48 +565,44 @@ public void visit( InputPortInfo n ) {
...
-				protocolVal = protocolExpr.evaluate();
-			}
-
-			if( protocolVal == null ) {
+			OLSyntaxNode protocolIdNode =
+				ModuleSystemUtil.transformProtocolId( n.protocolId(), interpreter.commCore() );
+			Expression protocolExpr = buildExpression( n.protocol() );
+			Expression protocolIdExpr = buildExpression( protocolIdNode );
+			protocolProcs.add( new DeepCopyProcess( protocolPath, protocolExpr, true, n.context() ) );
+			protocolProcs.add( new AssignmentProcess( protocolPath, protocolIdExpr, n.context() ) );
```
kicito, line 588: Here, we use ModuleSystemUtil.transformProtocolId to decide whether if we should transform a VariablePathExpressionNode which defines protocolId, into ConstantString or not, by determine if the identifier is supported by Jolie or not.

libjolie/src/main/java/jolie/lang/parse/ast/InputPortInfo.java
fmontesi: I think what you had was mostly fine. What I meant was simply that client code interested in getting the protocol id shouldn't get an entire OLSyntaxNode and then use `toString()`. That logic should be put inside of the method `protocolId()`. So you should have both your `protocol()` and the `protocolId()` methods. Does that make sense?

libjolie/src/main/java/jolie/lang/parse/ast/InputPortInfo.java
```diff
@@ -77,20 +75,35 @@ public InputPortInfo(
...
+			// a non optimized protocol node
+			OLSyntaxNode prodVal = ((SumExpressionNode) protocol).operands().iterator().next().value();
+			p = ((ProductExpressionNode) prodVal).operands().iterator().next().value();
+		}
+
+		if( p instanceof InlineTreeExpressionNode ) {
+			return ((InlineTreeExpressionNode) p)
+				.rootExpression().toString();
+		} else {
+			return p.toString();
```
fmontesi, line 106: Doesn't this risk throwing a nullpointerexception?

libjolie/src/main/java/jolie/lang/parse/ast/OutputPortInfo.java
```diff
@@ -52,15 +49,29 @@ public void accept( OLVisitor visitor ) {
...
+	public OLSyntaxNode protocolId() {
+		OLSyntaxNode p = this.protocol;
+		if( p instanceof SumExpressionNode ) {
+			// a non optimized protocol node
+			OLSyntaxNode prodNode = ((SumExpressionNode) protocol).operands().iterator().next().value();
+			p = ((ProductExpressionNode) prodNode).operands().iterator().next().value();
+		}
+
+		if( p instanceof InlineTreeExpressionNode ) {
+			return ((InlineTreeExpressionNode) p).rootExpression();
```
fmontesi, line 68: This is different than the one in `InputPortInfo`. Why? Should they?
If not, you might wanna extract this code to a static method in `PortInfo` and call it from both this method and the one in InputPortInfo.

javaServices/coreJavaServices/src/main/java/joliex/lang/inspector/Inspector.java
```diff
@@ -237,8 +237,8 @@ private static Value buildPortInfo( InputPortInfo portInfo, ProgramInspector ins
 		if( portInfo.location() != null ) {
 			result.setFirstChild( PortInfoType.LOCATION, portInfo.location().toString() );
 		}
-		if( portInfo.protocolId() != null ) {
-			result.setFirstChild( PortInfoType.PROTOCOL, portInfo.protocolId() );
+		if( portInfo.protocol() != null ) {
+			result.setFirstChild( PortInfoType.PROTOCOL, portInfo.protocol().toString() );
```
fmontesi, line 241: Should this just become protocolId like the rest?

javaServices/coreJavaServices/src/main/java/joliex/lang/inspector/Inspector.java
```diff
@@ -277,8 +277,8 @@ private static Value buildPortInfo( OutputPortInfo portInfo, Set< String > refer
 		if( portInfo.location() != null ) {
 			result.setFirstChild( PortInfoType.LOCATION, portInfo.location().toString() );
 		}
-		if( portInfo.protocolId() != null ) {
-			result.setFirstChild( PortInfoType.PROTOCOL, portInfo.protocolId() );
+		if( portInfo.protocol() != null ) {
```
fmontesi, line 280: Should this just become protocolId like the rest?

PR #254, opened 2020-06-30
Http hearders wildcard

bmaschio:
@fmontesi 

I have added the wild card to map all incoming Http headers to a Jolie specific node  

``` jolie
inputPort TestHttp {
    Location:"socket://localhost:80"
    Protocol: http{
        .headers.("*")= "headersFromWild";
        .default = "default"

    } 
    Interfaces: TestInterface
}
```

the result is 

```
                                               Value:
                                                        .data[0]
                                                        .userAgent[0] = PostmanRuntime/7.26.1 : string
                                                        .requestUri[0] = / : string
                                                        .headersFromWild[0]
                                                                .postman-token[0] = 3acc99d6-9364-4d9e-9259-5514ecbd62d6 : string
                                                                .host[0] = localhost : string
                                                                .connection[0] = keep-alive : string
                                                                .cache-control[0] = no-cache : string
                                                                .accept-encoding[0] = gzip, deflate, br : string
                                                                .user-agent[0] = PostmanRuntime/7.26.1 : string
                                                                .accept[0] = */* : string
                                                        .operation[0] =  : string
                                                        .cookies[0]
```


extensions/http/src/main/java/jolie/net/http/HttpMessage.java
```diff
@@ -178,6 +178,10 @@ public String getPropertyOrEmptyString( String name ) {
 		return (ret == null) ? "" : ret;
 	}
 
+	public Map< String, String > getProperties() {
```
fmontesi, line 181: Rename to `properties()`.

tools/jolie2openapi/pom.xml
```diff
@@ -0,0 +1,81 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
fmontesi, line 1: This file should not be included in this pull request.

tools/jolie2openapi/src/main/java/joliex/openapi/impl/support/RestDescriptor.java
```diff
@@ -0,0 +1,78 @@
+package joliex.openapi.impl.support;
```
fmontesi, line 1: This file should not be included in this pull request.

PR #261, opened 2020-08-05
Joliemock

klag:
Replaced jolie2summy (coded in Java) with joliemock (coded in jolie)


klag:
> Looks like the new test for joliemock is failing due to an md5 problem.

solved

PR #265, opened 2020-08-28
Fix missing type during initializing an internal service node.

kicito:
By add type definition from parent program to the internal service node.


fmontesi:
Thank you! :+1:

PR #270, opened 2020-09-04
Commandline argument for specify an execution service and logic for service target selection

kicito:
When executing `jolie hello.ol`. The following cases are considered.
1. if there is exactly one service defined in hello.ol, that's the one we launch for the target execution.
(we ignore the imported services)
2. if there are multiple services defined in hello.ol, then there is no default, the programmer MUST specify the service name.

Specifying a target service can be done through command line flag --service or -s for short. e.g.
```
jolie --service HelloService hello.ol
```

Where HelloService is one of the defined services in hello.ol.

### DIFF
--- a/jolie/src/main/java/jolie/Interpreter.java
+++ b/jolie/src/main/java/jolie/Interpreter.java
@@ -887,7 +887,7 @@ public class Interpreter {
 			tracer = new DummyTracer();
 		}
 
-		LOGGER.setLevel( Level.ALL );
+		LOGGER.setLevel( configuration.logLevel() );
 
 		exitingLock = new ReentrantLock();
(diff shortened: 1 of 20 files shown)


jolie/src/main/java/jolie/Interpreter.java
```diff
@@ -887,7 +887,7 @@ public Interpreter( Configuration configuration,
 			tracer = new DummyTracer();
 		}
 
-		LOGGER.setLevel( configuration.logLevel() );
+		LOGGER.setLevel( Level.ALL );
```
fmontesi, line 890: Why?

fmontesi (changes requested):

fmontesi:
Great, thanks! :+1:

PR #277, opened 2020-09-29
optimize link type symbol resolution

kicito:
Resolving a link type symbol to the root of the linking chain, instead of resolving to one level upper.


libjolie/src/main/java/jolie/lang/parse/module/SymbolReferenceResolver.java
```diff
@@ -567,6 +567,9 @@ public void visit( TypeDefinitionLink n ) {
 					return;
 				}
 				linkedType = (TypeDefinition) targetSymbolInfo.get().node();
+				while( linkedType instanceof TypeDefinitionLink ) {
```
fmontesi, line 570: Don't we risk that this loops infinitely in cases like the following? If so, we might fix this by putting an upper bound (100?).

```
type A:B
type B:A
```

fmontesi:
Looks great now, thank you very much. :100:

PR #282, opened 2020-10-21
Remove embed type check in EmbeddedServiceLoader

kicito:
As the previous commit has enabled service parameter type checking in the initialization of the interpreter. This PR eliminates the checking point on the embed statement "load" method. Also removes the requirement of the Type object to create a Service loader. So the upcoming feature of embedding service node with Module System in runtime service can create EmbeddedServiceLoader object without bother building an accepting parameter Type object.


fmontesi:
Thanks, looks good! :+1:

PR #286, opened 2020-12-01
Add support for stateless visitors of Jolie AST

xpicox:
### Introduced changes
1. `OLVisitor< C, R >` is a generic interface for any visitor of `OLSyntaxNode`. `visit` methods accept an additional argument (a context) of type `C` and return a value of type `R`, making it possible to define *stateless* visitors of the AST.

2. `VoidOLVisitor` extends `OLVisitor< Unit, Unit >` to provide the same interface as the *old* `OLVisitor`. Every visitor now implements `VoidOLVisitor`.


fmontesi (changes requested):
VoidOLVisitor -> UnitOLVisitor
instead of returning null in UnitOLVisitor, return Unit.instance.

fmontesi:
Awesome, thanks! :+1:

PR #287, opened 2020-12-12
adding Patch Verb

bmaschio:
@fmontesi 
I have implemented the PATCH verb changes I have tasted it with the attached file 
[TestPatch.zip](https://github.com/jolie/jolie/files/5683104/TestPatch.zip)


mwallnoefer:
Looks good, but CI failed?

bmaschio:
@mawolf  the interesting part the MVN install does not fail on my machines

fmontesi:
It's because maven couldn't fetch a package, not our problem. It's annoying when it happens, though...
Also, it looks like our test pom doesn't have ws-test as dependency? I get errors locally when compiling in parallel. @bmaschio @thesave can you look into that?

PR #298, opened 2021-02-26
added jvm memory data into stats operation response

bmaschio:


include/runtime.iol
```diff
@@ -108,6 +108,12 @@ type Stats:void {
 		.systemLoadAverage:double //< System load average
 		.version:string //< OS version
 	}
+
+	.jvm:void{
```
fmontesi, line 112: `jvm` is not very intuitive for me. What about `memory`? That way we don't need the `Mem` suffix in the subnodes.

```
memory {
  free:long
  used:long
  total:long
}
```

packages/runtime.ol
```diff
@@ -108,6 +108,11 @@ type Stats:void {
 		.systemLoadAverage:double //< System load average
 		.version:string //< OS version
 	}
+	.jvm:void{
```
fmontesi, line 111: See first comment.

javaServices/coreJavaServices/src/main/java/joliex/lang/RuntimeService.java
```diff
@@ -325,4 +326,11 @@ private void stats_files( Value stats ) {
 			stats.setFirstChild( "maxCount", unixBean.getMaxFileDescriptorCount() );
 		}
 	}
+
+	private void stats_jvm( Value stats ) {
+		Runtime runtime = Runtime.getRuntime();
```
fmontesi, line 331: See first comment.

fmontesi (changes requested):

include/runtime.iol
balintPixis: the idea was to be able to get more JVM related info in a future but I can change it not a problem

fmontesi (approved):

PR #299, opened 2021-03-03
Fix soap

thesave:
The PR fixes the dependencies for the SOAP extension (now we download them from maven and store them under `lib/jaxws`).
The PR also fixes the dependencies for the WS-Test package, used to test the SOAP extension (we also download its dependencies under the `lib` folder of the project and we load them from the Jolie procedure for the WS-Test).

The test now passes successfully ✅.

Unfortunately, I could not find a replacement for the package `com.sun.xml.ws`, which generates the warning below
— if you have any leads on this, please let me know :)

```WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.sun.xml.ws.model.Injector (file:.m2/repository/com/sun/xml/ws/rt/2.3.3/rt-2.3.3.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int)
WARNING: Please consider reporting this to the maintainers of com.sun.xml.ws.model.Injector
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```


fmontesi:
Great work, thanks! :)

PR #306, opened 2021-05-03
Function calls in jolie

krygergo:
### **Idea**
What we want to obtain is to have one or more function calls inside expressions. It is done by creating a specific OLSyntaxNode for these expressions called SolicitResponseExpression.
When the OLParser parses an expression, and it reads a sequence tokens of type `operationId@outputPortId( ... )` it will parse an SolicitResponseExpression for the AST.

It will the run the AST through the FunctionTranslator which will translate every SolicitResponseExpression, into a SolicitResponseOperationStatement. E.g:
`variable = operationId@outputPortId( input )`

Will translate into
`operationId@outputPortId( input )( #0 )`
`variable = #0`

### **Implementation**
The OLParser parsefactor method is changed, to when the parser meets a token is of type ID, it checks the next token for an AT sign.
If the next token is not an AT sign, it proceeds as normal by assigning the VariablePathNode path to the parsed variable.
If the next token is an AT sign, it creates and returns an SolicitResponseExpression containing the context, operationId, outputPortId and the parsed operation expression node.

The translation will then take place after the semantic verification inside the Interpreter so the sequence will look like OLParser -> OLParseTreeOptimizer -> SemanticVerifier -> FunctionTranslator.

When the FunctionTranslator is reached, the AST will be optimized and verified the focus is then on making the correct sequence between, the stored SolicitResponseExpression and the SolicitResponseOperationStatements with their associated VariablePathNodes.

### The FunctionTranslator
The TranslationVisitor class of the FunctionTranslator is of type OLVisitor< Unit, OLSyntaxNode >.

The logic behind the TranslationVisitor is, when it visits an SolicitResponseExpression. It maps the parsed SolicitResponseExpression to a new SolicitResponseOperationStatement and stores it in a global list. The visit then returns a VariablePathNode which contains the same hidden variable that was mapped to the SolicitResponseOperationStatement. The VariablePathNode then takes the place of the SolicitResponseExpression in the AST.

The visits where a sequence is not possible it only recurses on the containing OLSyntaxNode/Nodes. E.g CompareConditionNode cannot be a subNode of a SequenceStatement, therefore it would not make sense, to check if a SolicitResponseOperationStatement has to be added before a CompareConditionNode.

A visit where we want to check is inside the AssignStatement or SolicitResponseOperationStatement because it would make sense to have these statements as subNodes of a SequenceStatement, which is shown in the idea section.

The thing is we cannot only do this inside a SequenceStatement because there might not exist an SequenceStatement because of the OLParseTreeOptimizer. E.g
`main {`
`        println@Console( x@op( input ) )( )`
`}`
The AST would be DefinitionNode -> SolicitResponseOperationStatement -> SolicitResponseExpression. Which means it has to create a SequenceStatement on its own, for it to work as intended.

The AST after translation would be DefinitionNode -> SequenceStatement: children { SolicitResponseOperationStatement,  SolicitResponseOperationStatement }.
The first SolicitResponseOperationStatement will be `x@op( input )( #0 )`, the second will be `println@Console( #0 )( )`

It does so by checking if the list, containing SolicitResponseOperationStatements is not empty, when it returns after visiting the expression. If the list is not empty, it means in the subNodes of the statement, there exists a VariablePathNode which maps to a SolicitResponseOperationStatement in the list.
Then a SequenceStatement is created with the list, the visit statement is appended to the sequence and the visit returns the sequence. 

If a SequenceStatement allready exists it appends the children of the newly created SequenceStatement. This is fine because we know the sequence of the newly created SequenceStatement is in correct order. We don't want an AST with SequenceStatement -> SequenceStatement.


mwallnoefer:
Anybody has an opinion about this? I guess that the return value syntax had been excluded since the inception since it somehow contrasts the request-response (input-output) paradigm.

fmontesi:
No comments for now. There should be incoming PRs that do something similar but with the precious experience that has been gained here soon. :)

PR #314, opened 2021-05-19
Type.simplename() to fix bug in the inspector

kicito:
This PR fixes the issue on MetaJolie when inspecting the name of the type which was prefixed during parsing. By implementing type.simpleName() method that will return human-friendly type name to the client. This PR also removes type.id(), which is replaced by type.name().


fmontesi:
Thank you! :raised_hands:

klag:
Thank you!

PR #315, opened 2021-05-19
Import main.ol by default

kicito:
Module lookup for `main.ol` when the import path is a directory. 

The ModuleFinder will perform a lookup to a.b.c.main when the resolved path for a.b.c is a directory.

eg. 
```
from a.b.c import smt // resolves to a.b.c.main if a.b.c is a directory
```


fmontesi:
Thank you! :raised_hands: 

Can you add a note about this behaviour to https://docs.jolie-lang.org/v1.10.x/language-tools-and-standard-library/basics/modules.html ?

PR #326, opened 2021-06-10
added test for global array

SuperMauro1966:
Hi, if you define an array as global, when you try to access the elements wit for(el in array) , el is void and not assigned to the value of the element in the array.


fmontesi:
I've refactored the test to use the new test template in `primitives/test-template.ol`, please use that one in the future.

Can you add your own copyright line in the header? Currently only mine is shown, yours should be there as well since it's a new file.

I reproduced the bug. I'll create a branch fix/global-array where we can work once you add your copyright line.

SuperMauro1966:
@fmontesi 
I have inserted my name in the copyright.
I hope you are now enabled to continue your work.

fmontesi:
Perfect, thanks! :+1:

PR #328, opened 2021-06-11
bug fix in type extension utils

fmontesi:
Former-commit-id: d227bd5923896a56af91282ae847ee9dded98b4f


mwallnoefer (approved):
This is just a simple bug fix, so I think that this may be merged?

fmontesi:
I'm waiting for @klag to confirm this. We talked about it on discord and he didn't remember if this is still a fix and/or necessary.
Have you checked that it does the right thing perhaps, @mwallnoefer?

klag:
It seems that the current file is already correct, or am I wrong?

PR #339, opened 2021-07-08
fix Windows jolie2plasma launcher

mwallnoefer:


mwallnoefer:
@fmontesi this one too...

PR #340, opened 2021-07-08
fix converter tests on Windows

mwallnoefer:
pass the correct encoding parameter for the input files


mwallnoefer:
Also this may be merged, @fmontesi.

fmontesi:
Great, thanks! :)

PR #344, opened 2021-07-17
correct build send_encodeContent to send no content with delete http

bmaschio:
@fmontesi @mwallnoefer this is a simple change to align the delete behaviour with the get behaviour

### DIFF
--- a/extensions/http/src/main/java/jolie/net/HttpProtocol.java
+++ b/extensions/http/src/main/java/jolie/net/HttpProtocol.java
@@ -365,7 +365,8 @@ public class HttpProtocol extends CommProtocol implements HttpUtils.HttpProtocol
 						.append( '=' )
 						.append( URLEncoder.encode( v.strValue(), HttpUtils.URL_DECODER_ENC ) );
 
-					if( vecIt.hasNext() || nodesIt.hasNext() ) {
+					if( vecIt.hasNext() |
+					   nodesIt.hasNext() ) {
 						headerBuilder.append( '&' );
 					}
 				}
@@ -445,8 +446,8 @@ public class HttpProtocol extends CommProtocol implements HttpUtils.HttpProtocol
 	private EncodedContent send_encodeContent( CommMessage message, Method method, String charset, String format )
 		throws IOException {
 		EncodedContent ret = new EncodedContent();
-		if( inInputPort == false && (method == Method.GET | method == Method.DELETE) ) {
-			// We are building a GET request
+		if( inInputPort == false && (method == Method.GET || method == Method.DELETE) ) {
+			// We are building a GET or DELETE request
 			return ret;
 		}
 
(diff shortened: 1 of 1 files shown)


mwallnoefer (approved):
Perfectly okay, @bmaschio.

https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.5 states:
> A payload within a DELETE request message has no defined semantics;
> sending a payload body on a DELETE request might cause some existing
> implementations to reject the request.

klag:
It seems that DELETE may contain a body, so I think we have to pay attention to this fix.

mwallnoefer:
> It seems that DELETE may contain a body, so I think we have to pay attention to this fix.

@klag if we consider the mentioned RFC we would not be conforming sending a _request_ payload (not the response payload), but there could exist SW implementations which violate the norm. According to your answer I guess that you know about them, so maybe please give us an example.

klag:
The RFC does not state that the request body of the delete request must be empty, so I think it is better to fix this problem in another way. For example we could introduce a paramenter in the http protocol, which allows to define if the body must be included or not in a request instead of forcing it at the implementation level

bmaschio:
@klag I do not like the idea of introducing an extra parameter to configure something that by common use is expected to do not contain a body 
if we add a noBodyInDelete parameter then we need to added it also noBodyInGet giver to the fact that in the same standard is stated 
"   A payload within a GET request message has no defined semantics;
   sending a payload body on a GET request might cause some existing
   implementations to reject the request."

now in the current implementation  
```java

		if( inInputPort == false && method == Method.GET ) {
			// We are building a GET request
			return ret;
		}
```
the facto for an output port performing a get we send no body

My modification is just to align the behaviour

klag:
we could leave the empty body as a default, but I think it is important to give the possibility to insert a body, because it is not forbidden.

bmaschio:
@klag have you got an example of a real case where the delete has a body?

klag:
When you have to put a filter for selecting the resource. In some cases, when the number of the filter fields is so high, it could be not suitable to put them into the url as query parameters. You could put them inside the body instead.

bmaschio:
"The HTTP protocol does not place any a priori limit on the length of a URI. Servers MUST be able to handle the URI of any resource they serve, and SHOULD be able to handle URIs of unbounded length if they provide GET-based forms that could generate such URIs. A server SHOULD return 414 (Request-URI Too Long) status if a URI is longer than the server can handle (see section 10.4.15)."  The limit is given by the browser you use https://stackoverflow.com/questions/1051485/what-is-the-character-limit-on-url

klag:
I know, but, in any case, there is the possibility to design the fileds into the body. It is a possibility.

bmaschio:
@klag ok I think it is a real corner case I do not see the strict need to have it but if @fmontesi and @mwallnoefer agree with you I can implement the change with the extra parameter that I have no idea how to call it

klag:
@bmaschio maybe we could leave your fix, but we can add an enhancement in the issue list for fixing at the first useful opportunity

bmaschio:
Yes that could be an idea it will be 2 hours modification at the end

mwallnoefer:
@fmontesi are you able to merge this change?

fmontesi:
Yes, thanks for the good discussion. :)

fmontesi:
By the way, I've applied a small fix (see commit).

PR #348, opened 2021-09-14
fix add/eat token when access modifier token is present

kicito:
Fix incorrect behavior while parsing access modifier

During parsing the symbols, OLParser is incorrectly add/eat token when access modifier is present. Which results in an unexpected token exception when running this example code.

```jolie
from console import Console

private service ConfigurationService {
	inputPort IP {
		location: "local"
		protocol: p.protocol
		requestResponse: getDBConn( void )(string)
	}
	
	main {
		getDBConn ( req )( result ) {
			result = "SUPER_SECRET_CONN"
		}
	}
}

service MainService {

    embed Console as C
	embed ConfigurationService as Conf

	main {
		getDBConn@Conf()(res)
                 println@C(res)()
	}
}
```

### DIFF
--- a/libjolie/src/main/java/jolie/lang/parse/OLParser.java
+++ b/libjolie/src/main/java/jolie/lang/parse/OLParser.java
@@ -289,25 +289,20 @@ public class OLParser extends AbstractParser {
 
 	private void parseTypes()
 		throws IOException, ParserException {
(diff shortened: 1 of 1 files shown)


fmontesi:
Great, thanks! :+1:

PR #350, opened 2021-09-21
Service File: get the name of the file from which a service has been launched

xpicox:
This PR extends the API of the File service with two operations:
- getServiceFileName(void)(string)
- getRealServiceFileName(void)(string)

This is particularly useful for Jolie scripts (Jolie files with the shebang pattern `#!/usr/bin/env jolie`) which must retrieve the name of the file they are defined in, in order to provide good error messages.


fmontesi:
Thanks :)

PR #353, opened 2021-09-30
Call the right Jolie executor on OS Windows

mwallnoefer:


.idea/compiler.xml
```diff
@@ -66,7 +65,7 @@
       <module name="joliec" target="1.8" />
       <module name="jsonrpc" target="1.8" />
       <module name="launchers" target="1.5" />
-      <module name="libjolie" target="11" />
+      <module name="libjolie" target="1.8" />
```
fmontesi, line 68: Not sure about these changes. I remember at least that we're using APIs in `Optional` that require Java 11. It's very strange that this works... perhaps target is ignored?

.idea/compiler.xml
```diff
@@ -50,9 +50,8 @@
       <module name="http" target="1.8" />
       <module name="https" target="1.8" />
       <module name="javascript" target="1.8" />
-      <module name="jolie" target="11" />
+      <module name="jolie" target="1.8" />
       <module name="jolie-cli" target="1.8" />
-      <module name="jolie-distribution" target="1.5" />
```
fmontesi, line 55: Not sure where this removal comes from?

.idea/compiler.xml
```diff
@@ -50,9 +50,8 @@
       <module name="http" target="1.8" />
       <module name="https" target="1.8" />
       <module name="javascript" target="1.8" />
-      <module name="jolie" target="11" />
+      <module name="jolie" target="1.8" />
```
fmontesi, line 53: See other comment about Java 11/1.8.

mwallnoefer:
Okay, so let us do a partial revert of https://github.com/jolie/jolie/commit/bc2ab24978380bea5b908e248fbd4e98b17bef99. The problem was that you mixed two independent changes in one commit 😄 !

fmontesi:
> Okay, so let us do a partial revert of [bc2ab24](https://github.com/jolie/jolie/commit/bc2ab24978380bea5b908e248fbd4e98b17bef99). The problem was that you mixed two independent changes in one commit 😄 !

You're right! Thanks for the extra work. 😅

PR #356, opened 2021-10-04
Fixes for running exec tests

fmontesi:
Might be necessary to access the Jolie executable in github actions.


fmontesi:
Still a WIP. :)

PR #374, opened 2022-03-28
Introducted logging for http/soap protocols.

SuperMauro1966:
If monitoring is active, I have introducted the http / soap logging.

Please note that because inside HttpPrococol the recv_internal ha changed the return type, this has caused a domino effect on many classes.

So there is a lot of changes in the project concerning the return type: not anymore CommMessage but CommMessageProtocol.

Sorry, I often did commit with italian comments forgotting the international environment of the project.


fmontesi:
Thanks! I'd really like to have this feature and the monitoring code you added. I'd just like not to change the protocol/messaging API, I've kickstarted the discussion on how to do that on discord. Let's reopen this once we converge there.

PR #378, opened 2022-03-31
startServiceSession - added missing operation start log in case execution = sequential

SuperMauro1966:
In the above operation, the sequential case didn't log the operation start event.


kicito (approved):
LGTM :)

PR #381, opened 2022-04-20
Features/monitoring http

SuperMauro1966:
@fmontesi 
A lot of modifications are in this pull request.

I introduced the rawId field inside a CommMessage, following out previous talk with @klag .
So when we are going to fire an MessageEvent we send now a rawId field to the Monitor.

In send_internal, recv_internal (http and soap) there is now a fire event for ProtocolMessage. 

I tried to be careful in modifying the code.

If something is not clear please write me, I tried to explain my ideas.


jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
```diff
@@ -0,0 +1,74 @@
+package jolie.monitoring.events;
```
fmontesi, line 1: Missing license header.

extensions/http/src/main/java/jolie/net/HttpProtocol.java
```diff
@@ -1201,27 +1220,49 @@ private static void recv_parseJsonQueryString( HttpMessage message, Value value,
...
-				.append( new String( message.content(), charset ) );
-		}
-		return debugSB.toString();
+		return headerStr.toString();
+	}
+
+	/*
+	 * Prints debug information about a received message
+	 */
+	private String getHttpBody( HttpMessage message, String charset )
```
fmontesi, line 1257: Could be static?

extensions/http/src/main/java/jolie/net/HttpProtocol.java
```diff
@@ -1201,27 +1220,49 @@ private static void recv_parseJsonQueryString( HttpMessage message, Value value,
...
+			debugSB.append( "--> Message content\n" )
+				.append( getHttpBody( message, charset ) );
+		}
+		return debugSB.toString();
+	}
+
+	/*
+	 * return the received message's header
+	 */
+	private String getHttpHeader( HttpMessage message )
```
fmontesi, line 1234: Could be static perhaps?

extensions/soap/src/main/java/jolie/net/SoapProtocol.java
```diff
@@ -697,6 +700,7 @@ public void send_internal( OutputStream ostream, CommMessage message, InputStrea
 
 		final StringBuilder httpMessage = new StringBuilder();
 		ByteArray content = null;
+		String bodyMessage = "???"; // body message for ProtocolMessageEvent
```
fmontesi, line 703: Why "???" and not just empty ""?

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
```diff
@@ -0,0 +1,74 @@
...
+		PROTOCOL( "protocol" ), MESSAGE( "message" ), HEADER( "header" ), BODY( "body" ), PROCESSID(
+			"processId" ), RAWID( "rawId" );
+
+		private String fieldName;
+
+		FieldNames( String name ) {
+			this.fieldName = name;
+		}
+
+		public String getName() {
```
fmontesi, line 32: `name()`

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
```diff
@@ -0,0 +1,74 @@
...
+		Protocol( String protocol ) {
+			this.protocol = protocol;
+		}
+
+		public String getProtocol() {
+			return protocol;
+		}
+	}
+
+	public static enum FieldNames {
```
fmontesi, line 22: `Field`

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
```diff
@@ -0,0 +1,74 @@
...
+
+import jolie.monitoring.MonitoringEvent;
+import jolie.runtime.Value;
+
+public class ProtocolMessageEvent extends MonitoringEvent {
+
+	public enum Protocol {
+		SOAP( "soap" ), HTTP( "http" );
+
+		private String protocol;
```
fmontesi, line 11: `final`

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
```diff
@@ -0,0 +1,74 @@
...
+		public String getProtocol() {
+			return protocol;
+		}
+	}
+
+	public static enum FieldNames {
+		PROTOCOL( "protocol" ), MESSAGE( "message" ), HEADER( "header" ), BODY( "body" ), PROCESSID(
+			"processId" ), RAWID( "rawId" );
+
+		private String fieldName;
```
fmontesi, line 26: `final`

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
```diff
@@ -0,0 +1,74 @@
...
+	public enum Protocol {
+		SOAP( "soap" ), HTTP( "http" );
+
+		private String protocol;
+
+		Protocol( String protocol ) {
+			this.protocol = protocol;
+		}
+
+		public String getProtocol() {
```
fmontesi, line 17: `name()`

jolie/src/main/java/jolie/monitoring/events/OperationCallEvent.java
```diff
@@ -34,7 +34,7 @@ public class OperationCallEvent extends MonitoringEvent {
 
 
 	public OperationCallEvent( String operationName, String processId, String messageId, int status, String details,
-		String outputPort, Value message ) {
+		String outputPort, Value message, String rawId ) {
```
fmontesi, line 37: `rawId` is not descriptive. Could you suggest a better name? `messageId` should probably become `requestId` and `rawId` might become `messageId`.

fmontesi: (Same comment for the Jolie types corresponding to this.)

fmontesi (changes requested):

jolie/src/main/java/jolie/monitoring/events/OperationCallEvent.java
SuperMauro1966: @fmontesi @klag 
I used the convention decided in our last conversation.
In the db  associated with Monitor, I also created the same identifier, and rawId is userd in isJolieLoggingMonitor.

Next time perhaps is better working with a written analysis.

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
SuperMauro1966: [ERROR] /mnt/d/my-jolie-fork/jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java:[54,31] name() in jolie.monitoring.events.ProtocolMessageEvent.Field cannot override name() in java.lang.Enum

choosed to call fieldName

extensions/soap/src/main/java/jolie/net/SoapProtocol.java
SuperMauro1966: The ratio was to signal an anomaly in retriving, in the execution of the code, the body text.
The bodyMessage is setted inside a try ... catch and peraphs an error situation managed by the catch could be throw before the bodyMessage retrieval.

I suggest this change

String bodyMessage = "--- unable to retrieve the body data ---";

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
fmontesi: Good point. Then maybe `id()`?

fmontesi: (PS: the reason is that we do not use the `get` prefix in getters in the Jolie codebase.)

jolie/src/main/java/jolie/monitoring/events/OperationCallEvent.java
fmontesi: I do not recall saying that `rawId` would be fine (and I'd be surprised at me if I did). Sorry if I did. Any proposal for alternative? I think that `rawId` is going to be confusing to you as well going forward. `raw` is misleading. We have internal message ids (what you call rawId) and request ids.

SuperMauro1966: Hi do you think protocolMessageId would be nice?

klag: @fmontesi could we decide the name of the field as soon as possible? It has impacts on some code on our parts

extensions/http/src/main/java/jolie/net/HttpProtocol.java
SuperMauro1966: Changed as requested.

extensions/http/src/main/java/jolie/net/HttpProtocol.java
SuperMauro1966: Changed as requested.

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
SuperMauro1966: enum FieldName changed to enum Field

hidden field named id,
access method to this field changed in id()

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
SuperMauro1966: Changed as requested.

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
SuperMauro1966: Changed in protocol.

extensions/soap/src/main/java/jolie/net/SoapProtocol.java
SuperMauro1966: removed bodyMessage.

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
SuperMauro1966: Changed as requested.

jolie/src/main/java/jolie/monitoring/events/ProtocolMessageEvent.java
SuperMauro1966: Changed as requested.

jolie/src/main/java/jolie/monitoring/events/OperationCallEvent.java
fmontesi: Mmh, `protocolMessageId` makes it seems related to the protocol, but every message has it regardless of the protocol.
What about making `messageId` the actual id of the message (instead of the request id as of now), and then a new `requestId` field that contains the request id? Would this be changing too much in existing implementations of monitors?

klag: Since messageId alerady exixts we should change two variables instead of one. 
I suggest to avoid to alter the current messageId leaving this modification for major revision in the future.

rawMessageId?

fmontesi: Sounds good, let's avoid changing too much in one step if it causes problems.
I suggest `internalMessageId` then. `raw` seems to be "low-level" but it's actually on the same level, just different.

PR #390, opened 2022-05-07
Function call inside expression

match19:
This commit adds a new feature to Jolie, which makes it possible to call functions inside expressions.

**Before:**
operationName@OutputPortName( param )( result )
result = result + 1

**Now:**
result = operationName@OutputPortName( param ) + 1

The test **primitives/solicit_response_expression.ol** is added to test this behavior.

### DIFF
--- a/jolie/src/main/java/jolie/OOITBuilder.java
+++ b/jolie/src/main/java/jolie/OOITBuilder.java
@@ -500,7 +500,11 @@ public class OOITBuilder implements UnitOLVisitor {
 				return;
 			}
 			if( protocolExpr instanceof Value || protocolExpr instanceof InlineTreeExpression ) {
-				protocolStr = protocolExpr.evaluate().strValue();
+				try {
+					protocolStr = protocolExpr.evaluate().strValue();
+				} catch( FaultException e ) {
+					throw new AssertionError( "Expression.evaluate() should never throw an exception here" );
+				}
 			} else if( protocolExpr instanceof VariablePath ) {
 				VariablePath path = new ClosedVariablePath( (VariablePath) protocolExpr, initValue );
 				protocolStr = path.getValue().strValue();
@@ -1617,6 +1621,20 @@ public class OOITBuilder implements UnitOLVisitor {
 			error( n.context(), e );
 		} catch( InvalidIdException e ) {
 			error( n.context(), "could not find port " + n.bindingPort().id() );
+		}
+	}
+
+	public void visit( SolicitResponseExpressionNode n ) {
+		try {
+			currExpression =
+				new SolicitResponseExpression(
+					n.id(),
+					interpreter.getOutputPort( n.outputPortId() ),
+					buildExpression( n.outputExpression() ),
+					solicitResponseTypes.get( n.outputPortId() ).get( n.id() ),
+					n.context() );
+		} catch( InvalidIdException e ) {
+			error( n.context(), e );
 		}
 	}
 }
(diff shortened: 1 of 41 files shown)


test/primitives/solicit_response_expression.ol
```diff
@@ -0,0 +1,40 @@
+from ..test-unit import TestUnitInterface
```
fmontesi, line 1: Missing license header.

test/primitives/solicit_response_expression.ol
```diff
@@ -0,0 +1,40 @@
...
+
+include "math.iol"
```
fmontesi, line 3: Include is deprecated, please use imports.

fmontesi (changes requested):

jolie/src/main/java/jolie/OOITBuilder.java
```diff
@@ -500,7 +500,11 @@ public void visit( InputPortInfo n ) {
 				return;
 			}
 			if( protocolExpr instanceof Value || protocolExpr instanceof InlineTreeExpression ) {
-				protocolStr = protocolExpr.evaluate().strValue();
+				try {
+					protocolStr = protocolExpr.evaluate().strValue();
+				} catch( FaultException e ) {
+					throw new AssertionError( "Expression.evaluate() should never throw an exception here" );
```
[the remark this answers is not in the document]
match19: When defining the protocol for example `protocol: http { format = "json" }`. Having function calls inside this definition is not support eg.  `protocol: http { format = trim@StringUtils(" json")}`, (Another error will occur). Because of this I assumed that it is not supposed to be possible that the `protocolExpr `could be a `SolicitResponseExpression`, and thus the FaultException should never happen.

test/primitives/solicit_response_expression.ol
```diff
@@ -0,0 +1,40 @@
...
+
+service Test {
+	inputPort TestUnitInput {
+		location: "local"
+		interfaces: TestUnitInterface
+	}
+
+    
+	main {
+		test()() {
```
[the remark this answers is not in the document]
match19: The indentation should be that same as in the `test-template.ol` file, I tried copy pasting again and saw no changes ?

fmontesi:
> This pull request **introduces 1 alert** when merging [2a91506](https://github.com/jolie/jolie/commit/2a915064ee305d9abe162f40a157fa6ec711ba28) into [f059b72](https://github.com/jolie/jolie/commit/f059b72120b7680d2cbb15193a246424bc0976c0) - [view on LGTM.com](https://lgtm.com/projects/g/jolie/jolie/rev/pr-ffbbe44451ae14ef6cc5efa69ecc5f5c943e6f61)
> 
> **new alerts:**
> 
> * 1 for Dereferenced variable may be null

This should be fixed.

fmontesi:
Lots of places now have to handle `FaultException`. Using `RuntimeFaultException` might help in simplifying the code. @thesave used it in the past to make expressions capable of throwing faults like `ArithmeticException`.

libjolie/src/main/java/jolie/lang/parse/OLParser.java
```diff
@@ -3303,13 +3314,19 @@ private OLSyntaxNode parseFactor()
 
 		switch( token.type() ) {
 		case ID:
+			String id = token.content();
 			path = parseVariablePath();
```
fmontesi, line 3318: Mmmh, what happens if one writes `x.y@Hello( "x" )`, which should be forbidden? It looks like you'll parse it incorrectly as x@Hello("x").

Issue #327, opened 2021-06-11
Navigation of global arrays


fmontesi:
**Describe the bug**
Navigating a global array with `for( .. in ..)` does not work as expected.

**To Reproduce**
1. Checkout `fix/global-array`
2. Run the test `primitives/global-array.ol`

**Expected behavior**
The test should pass.

**Desktop (please complete the following information):**
 - OS: Linux
 - Jolie version: 1.10.4
 - Java version: 11

kicito:
Hi, I am trying to fix this bug as it might be relevant to the project I am working on. As far as I tried to debug the test case. For this part of the code.

```
for( item in global.array ) {
	if( item != i ) {
		throw( TestFailed, "global.array[" + i + "] = " + item + "instead of " + i )
	}
	i++
}
```

The for ... in loop creates an additional instance of `ValueLink` and attached to the variable `item` but the evaluation of comparing `item (ValueLink)` with `i (ValueImpl)` fails. As shown in the image of my debugger.

![image](https://user-images.githubusercontent.com/9067906/171123721-753fea9f-25a3-410b-bdca-7831dad6bc18.png)

Now I am not sure if we should fix the for ... in process to bind the raw value inside the or the evaluation of the ValueLink, could you give me any hint @fmontesi ?

fmontesi:
Good catch. That's because `ValueLink` doesn't implement `equals`, so it defaults to object equality (which is clearly false in this case).

I don't know if using `equals` in there is a good/robust idea, since an implementation of equality for `ValueLink` will have to visit the linked value, which might be surprising. We should probably have a method called differently, like `isEqualTo( Value v )`, which is offered as abstract method by `Value` and is then implemented by subclasses. The implementation in `ValueLink` should invoke `getLinkedValue.isEqualTo`, I guess. Then we can call it from within Jolie. We should check if we call `equals` from anywhere in the code and replace it with a call to `isEqualTo`.

Does this make sense?

PR #401, opened 2022-06-03
Fix/global array

kicito:
fix according to discussion on #327 
This PR includes

1. Add an override copy method on GlobalVariablePath, as currently the information has lost during VariablePath copy method 
2. Add Value.isEqualTo method to replace calling to Value.equals


fmontesi:
I see this problem in the CI pipeline. Not sure it's related to this PR... opinions?

Screenshot_20220603_130518

PR #402, opened 2022-06-15
Fix incorrect jar name

kicito:
Which led to the ClassNotFound exception during the execution of the program.


kicito:
I see  jolie/release_tools, jolie/experiment, jolie/backup... that need to apply this change. Is there other place I need to lookup?

fmontesi:
No, that should be it. :) I was thinking of launchers/jolie2java etc., but
it looks like they are not using json anywhere..

PR #404, opened 2022-06-21
Module Finder algorithm for the package manager

kicito:
Change module finder algorithm to support future implementation of package manager

```
Given WDIR = current working directory P = importing tokens FIRST = first token of import path
REST = the rest of importing tokens
1. Check if FIRST.jap is in WDIR/lib. If so, resolve REST inside of it.
2. Try to resolve P from the packages directory, up one level, until system root 
3. Try to resolve P from the list of packages directories passing through -p flag
```

In practice: 
Given execution target main.ol located in /home/usr/example/, current working directory is /home/usr/example
an import statement from A import a will try to find A in the following location (order by precedence)
1. /home/usr/example/lib/A.jap ( $cwd/lib/fst_token.jap )
2. /home/usr/example/packages/A.ol
3. /home/usr/packages/A.ol
4. /home/packages/A.ol
5. /packages/A.ol
6. list of specific directory passing by -p flag ( default ${JOLIE_HOME}/packages )

Note: this change depends on `fix/parentInterpreter`


fmontesi:
Great work, thank you! I'll review it asap. :)

kicito:
Seems like the test is failed on the CI (works fine on my local machine XD)... I will have a look at it

extensions/jolie-embedding-legacy/src/main/java/jolie/embedding/jolie/JolieServiceLoader.java
```diff
@@ -58,22 +58,36 @@ public JolieServiceLoader( Expression channelDest, Interpreter currInterpreter,
...
-			config.connectionsLimit(), config.cellId(), config.correlationAlgorithm(), config.includePaths(),
-			config.optionArgs(),
-			config.libUrls(), config.inputStream(), config.charset(), config.programFilepath(), config.arguments(),
-			config.constants(),
-			config.jolieClassLoader(), config.isProgramCompiled(), config.typeCheck(), config.tracer(),
-			config.tracerLevel(),
-			config.tracerMode(), config.check(), config.printStackTraces(), config.responseTimeout(), config.logLevel(),
-			config.programDirectory(), config.packagePaths(),
+		Interpreter.Configuration config = Interpreter.Configuration.create(
+			commandLineParser.getInterpreterConfiguration().connectionsLimit(),
```
fmontesi, line 62: To shorten this out, could you put `commandLineParser.getInterpreterConfiguration()` in a temp variable (`cmdConfig`?) before these lines and then use that here?

libjolie/src/main/java/jolie/lang/parse/module/ModuleFinderImpl.java
```diff
@@ -137,6 +151,40 @@ private ModuleSource moduleLookup( Path basePath, ImportPath importPath ) throws
...
+	 * @return source object to be parsed by module parser.
+	 */
+	private ModuleSource moduleLookupFromPackages( Path basePath, ImportPath importPath )
+		throws FileNotFoundException {
+
+		Path targetPath = basePath;
+		String[] lookupPaths = new String[ basePath.getNameCount() ];
+		for( int i = 0; i < basePath.getNameCount(); i++ ) {
+			boolean shouldPopTwice = false;
+			if( !targetPath.getFileName().toString().equals( "packages" ) ) {
```
fmontesi, line 169: Let's make `packages` a constant in `jolie.Constants`

fmontesi (changes requested):

libjolie/src/main/java/jolie/lang/parse/module/Modules.java
```diff
@@ -36,13 +37,18 @@ public Map< URI, SymbolTable > symbolTables() {
...
 		ModuleParser parser = new ModuleParser( configuration );
-		ModuleFinder finder = new ModuleFinderImpl( configuration.packagePaths() );
-
 		ModuleRecord mainRecord = parser.parse(
-			new Scanner( stream, programDirectory, configuration.charset(), configuration.includeDocumentation() ) );
+			new Scanner( stream, programURI, configuration.charset(), configuration.includeDocumentation() ) );
+		ModuleFinder finder;
+
+		if( programURI.toString().contains( "jap:" ) ) {
+			finder = new ModuleFinderDummy();
```
kicito, line 48: I don't really like this solution. I will think about the alternative later... The reason behind this now we need to pass the module path to finder. But Windows complain about having : on the path `jap:file`, so I create this Dummy to bypass the module lookup phase. As for now we do not support such thing in `jap` yet.

fmontesi: I see! I'll add a todo to remember that we need to re-evaluate this one day.

PR #414, opened 2022-07-11
Add base64 to security utils

kicito:
Adds operation base64 to the security utils package.


fmontesi:
Do the operations in `converter` already cover this, perhaps? They take raw instead of string though.

PR #417, opened 2022-07-13
Rename all "Main" to use a constant in defined in ServiceNode

kicito:
This PR removes the redundancy use of the keyword "Main" vs "main" for `ServiceNode` related stuff.


fmontesi:
Shouldn't it be "Main"? Service definition names start with an uppercase letter. Is that what the constant represents?

fmontesi:
Or is it about the standard module name? Maybe I misinterpret the code.. 🙂

kicito:
Previously when parsing old Jolie syntax file, Jolie will transform into a Service named `Main`. But the default Jolie import target is pointing to `main`.  So here I tried to consolidate the values to point to one place. We can change to `Main` as well.

fmontesi:
Ah I see!

By convention services start with an uppercase letter and modules with a
lowercase letter.

So "Main" for the service and "main" for the module. 🙂

kicito:
Right, sorry for my confusion. Let me add some comments to the code on this PR. Please do look at it :)

fmontesi:
Looks good! Thanks. 🙂

PR #423, opened 2022-07-28
fix: side effect on handling querystrings

kicito:
I don't think this is a efficient way to fix the issue. But currently there are side effect on appending querystring before sending an HTTP request. To be concrete, `send_appendQuerystring` has side-effect that it removes value of the sending message. Then the methods that execute later, such as header setter (snippet below), wouldn't received the proper value as intended.

https://github.com/jolie/jolie/blob/960417451299cfc6631a36272c8866df5163aad2/extensions/http/src/main/java/jolie/net/HttpProtocol.java#L886-L890

### DIFF
--- a/extensions/http/src/main/java/jolie/net/HttpProtocol.java
+++ b/extensions/http/src/main/java/jolie/net/HttpProtocol.java
@@ -85,8 +85,10 @@ import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Set;
+import java.util.StringJoiner;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 /**
  * HTTP protocol implementation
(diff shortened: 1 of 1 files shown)


fmontesi:
An efficiency way of fixing it would be to modify send_appendQuerystring. Just create a Map copy of value.children() and iterate over that instead of value.children() in the rest of the method. That should work. :) Since it'd be a shallow copy (the map just contains references) it should be OK memory-wise.

Could you try doing that instead of cloning the value?

kicito:
Not exactly what you said, but I think this is better than the last one. I came up with this since `Map.copyOf` returns an unmodifiable Map so the old logic of iterate and remove doesn't work.

fmontesi:
Looks good to me! :)
I didn't mean Map.copyOf, more like new HashMap<>( originalMap ), but this works too. :)
That `List` might not be that efficient, a HashSet would probably be better, but it's good enough for now.

PR #424, opened 2022-08-02
Check IndexOutOfBound when building error message

kicito:
Running the following snippet
```
service main {
    main {
        nullProcess
}
```
prints `Index -1 out of bounds for length 0` on the terminal.

Note: there might be a case that is left unchecked. I haven't revised the whole file.


kicito:
Hmmm, CI works fine on my repo ... https://github.com/kicito/jolie/actions/runs/2782033243

PR #433, opened 2022-10-12
fix in/decrement process and add type precedence on evaluating equal …

### LINKED ISSUE #432: ++ and -- do not work with long
`    a = 1662929984215L
    a++
    println@Console( a )()    `
    
    the result is `777640662`

if we try
`    a = 1662929984215L
    a = a + 1
    println@Console( a )()  `

the result is `1662929984216`

kicito:
This PR fixes #432.  Also, we discover another related bug where Jolie always cast the value's type based on the left expression. Which breaks the symmetric property of equality as the code snippet below shows.  This PR fixes it as well :)

```
	if ( 777640662 == 1662929984216L ) { // cast both to int, do not throws
		throw( TestFailed, "expected 777640662 != 1662929984216L, but it is evaluated to true " + a )
	}
	
	if ( 1662929984216L == 777640662 ) {// cast both to long, throws
		throw( TestFailed, "expected 1662929984216L != 777640662, but it is evaluated to true " + a )
	}
```

### DIFF
--- a/jolie/src/main/java/jolie/process/PostDecrementProcess.java
+++ b/jolie/src/main/java/jolie/process/PostDecrementProcess.java
@@ -22,6 +22,8 @@
 package jolie.process;
 
 import jolie.ExecutionThread;
+import jolie.lang.Constants;
+import jolie.runtime.FaultException;
 import jolie.runtime.Value;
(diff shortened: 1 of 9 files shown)


fmontesi:
Excellent, thanks! :)

PR #445, opened 2022-12-21
corrected @AddJarDeps with jar without version

bmaschio:
I have removed the version from the dependency @AddJarDeps
@fmontesi I think we should add a test on the service that is currently not covered by it


fmontesi:
Great thank you! Can you add a test in another PR? :)

PR #449, opened 2023-02-20
Fix Cleaner deamon thread is not closing after the service termination

kicito:
Due to the reference from Listeners that keeps the interpreter object in the memory Cleaner daemon thread never terminates and stays alive throughout the whole execution. This PR clears the interpreter object and erases the state location value on the listener shutdown. It also uses WeakReference for the interpreter to help the gc clean the reference.

### DIFF
--- a/extensions/btl2cap/src/main/java/jolie/net/BTL2CapListener.java
+++ b/extensions/btl2cap/src/main/java/jolie/net/BTL2CapListener.java
@@ -73,12 +73,10 @@ public class BTL2CapListener extends CommListener {
 	}
 
 	@Override
-	public void shutdown() {
+	public void onShutdown() {
 		try {
 			connectionNotifier.close();
 		} catch( IOException e ) {
 		}
-		super.clearInterpreter();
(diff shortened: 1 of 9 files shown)


extensions/btl2cap/src/main/java/jolie/net/BTL2CapListener.java
```diff
@@ -78,5 +78,7 @@ public void shutdown() {
 			connectionNotifier.close();
 		} catch( IOException e ) {
 		}
+		super.clearInterpreter();
```
fmontesi, line 81: We do these two lines a bit repetitively.. Maybe they could be put in an abstract class that the relevant listeners inherit from, or maybe we could even use WeakReference for the interpreter references in the superclass and the location value?

kicito:
Updated the code, I already add WeakReference to the superclass :) the `clearInterpreter()` somehow is still needed when monitoring the threads on the testing program. But now the test failed on the ci (my machine works fine)... :(

PR #455, opened 2023-03-09
Fixed parsing context for some tokens

EmilOvcina:
The parsing context for ID and string tokens now has the correct start and end columns. 
This is used to get the correct parsing contexts for embeds, ports (and properties of ports), strings, and ID tokens.

The span for the contexts is chosen based on how types are parsed.

The span of the parsing context for embeds and ports is highlighted in **bold** and _italic_ here:
<code>emded _**SVC as PORT**_</code>

<code>inputPort _**IP { ... }**_ </code>

Strings now span over both quote characters. For example:
<code>Location: _**"socket://localhost:9000"**_</code>

ID tokens also span over _InlineTreeExpressions_. For example:
<code>Protocol: _**http { .format="json" }**_</code>


fmontesi:
Great fix, thanks! :)

PR #459, opened 2023-05-16
Add missing new line when printing empty arrays

lucat1:
All the branches of this method end up printing a new line (either via recursion or the base case of `writeNativeValue`), except in the case of an empty array. This patches adds the missing new line character.


fmontesi:
Thank you! :⁠-⁠)

PR #465, opened 2024-01-29
string-utils.ol - fix percentage test

mwallnoefer:
We need to provide the locale for the percentage test.


fmontesi:
Good catch, thank you Matthias! :-)

PR #466, opened 2024-02-28
HttpProtocol.java - fix return status code in case of an invalid code

mwallnoefer:
Don't pass the illegal status code onwards, we need to exchange it by the numeric value 500. The correct description gets set afterwards.


mwallnoefer:
Hi Fabrizio,

this patch was the preliminary attempt to fix some more-in-depth issue: at the moment it is not possible to set the `statusCode` as a constant, since then we always end up in the error branch (`getIntParameter()` returns 0).

In other words, 
`
	inputPort HTTPInput {
		protocol: http {
			statusCode -> statusCode...
`
is possible, but the following snippet does not work:
`
	inputPort HTTPInput {
		protocol: http {
			statusCode = statusCode // OR EVEN statusCode = "200"...
`

How can we do better?

PR #469, opened 2024-03-18
HttpProtocol.java - do not provide provide a payload for one-way responses and req/resp ones with return type void

mwallnoefer:
HTTP server one-way responses should not provide a payload and a content-type: 0 header. This behaviour is required for compatibility with certain HTTP operations as "204 No Content".

### DIFF
--- a/extensions/http/src/main/java/jolie/net/HttpProtocol.java
+++ b/extensions/http/src/main/java/jolie/net/HttpProtocol.java
@@ -392,7 +392,7 @@ public class HttpProtocol extends CommProtocol implements HttpUtils.HttpProtocol
 		}
 	}
 
-	private void send_appendJsonQueryString( CommMessage message, StringBuilder headerBuilder )
+	private void send_appendJsonQueryString( CommMessage message, Type sendType, StringBuilder headerBuilder )
 		throws IOException {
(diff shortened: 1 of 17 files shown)


mwallnoefer:
These two additional commits bring ReqResp void message responses to the same level like one-way messages. It seems to be a sane behaviour that allows us to send back also different error codes on HTTP like 204 without any payload.

extensions/http/src/main/java/jolie/net/HttpProtocol.java
```diff
@@ -681,7 +684,7 @@ && hasOperationSpecificParameter( message.operationName(), Parameters.STATUS_COD
 					" is sending a message with status code " +
 					statusCode +
 					", which is not in the HTTP specifications." );
-				statusCode = 500;
+				statusCode = DEFAULT_STATUS_CODE;
```
fmontesi, line 687: Isn't this an internal server error? (The server is doing something that doesn't make sense because it is misconfigured.)

fmontesi: Maybe we could add an exception to this method, throw it containing the new fault, and then catch it at an outer caller to change the comm message?

PR #477, opened 2024-03-27
HTTP protocol - generate proper exceptions on wrong status codes

mwallnoefer:
* do not bail out on send() I/O exceptions when the client could be notified
* refactoring of the internal status codes structures in HttpUtils


fmontesi:
Very good stuff, thank you Matthias! :-)

PR #478, opened 2024-03-28
Fix/http refactoring

mwallnoefer:


mwallnoefer:
Good catch. In _HttpProtocol.java_ I find the following call which seems to return `false` unless specified differently by means of a HTTP parameter.
```
	@Override
	public boolean isThreadSafe() {
		return checkBooleanParameter( HttpUtils.Parameters.CONCURRENT );
	}
```

So I guess that the following attributes are thread-dependent:
```
	private String inputId = null;
	private MultiPartFormDataParser multiPartFormDataParser = null;
	private String encoding = null;
	private String requestFormat = null;
	private String requestCharset = null;
	private boolean headRequest = false;
```
The most easy solution should be the use of the `ThreadLocal` helper class.

fmontesi:
Unfortunately you don't get any guarantee that a response is gonna be
served by the same thread that parsed the request, in fact they're probably
gonna be different.

We could change this model thanks to Java's new virtual threads, and we
probably should! But that's a bit longer down the lane...

Issue #388, opened 2022-05-04
Database Service attributes


bmaschio:
From the [DbService](https://github.com/jolie/jolie/blob/master/javaServices/coreJavaServices/src/main/java/joliex/db/DatabaseService.java)
it seems that the attributes are applied only in case of enbedding

```java
	if( isEmbedded ) {
		connectionString = "jdbc:" + driver + ":" + databaseName;
		if( !attributes.isEmpty() ) {
			connectionString += ";" + attributes;
		}
		if( "hsqldb".equals( driver ) ) {
			connection = DriverManager.getConnection(
				connectionString,
				username,
				password );
		} else {
			connection = DriverManager.getConnection( connectionString );
		}
	} else {
		if( driver.startsWith( "hsqldb" ) ) {
			connectionString = "jdbc:" + driver + ":" + driver.substring( driver.indexOf( '_' ) + 1 ) + "//"
				+ host + (port.isEmpty() ? "" : ":" + port) + separator + databaseName + "?characterEncoding="
				+ encoding;
		} else {
			connectionString =
				"jdbc:" + driver + "://" + host + (port.isEmpty() ? "" : ":" + port) + separator + databaseName
					+ "?characterEncoding=" + encoding;
		}
		connection = DriverManager.getConnection(
			connectionString,
			username,
			password );
	}
```
Am I wronge

mwallnoefer:
This seems to be the case.

Does PR #493 change something here? Otherwise @bmaschio would you like to propose some patch?

mwallnoefer:
PR #548

PR #493, opened 2024-04-08
Adding support for transactions to DatabaseService

Maje419:
In this PR, the DatabaseService is updated to support multiple open connections to the same database. This is done through a [HikariCP](https://github.com/brettwooldridge/HikariCP) connection pool.

I've strived to keep functionality exactly the same, to have complete backwards compatibility.

The following operations have been added:
**beginTx(void)(int):** This operation starts a transaction by removing one of the connections from the connection pool, and placing it into a ConcurrentHashMap for open connections. It creates and assigns this this transaction a transaction handle, which is returned and can be used to refer to this specific transaction in other calls.
**commitTx(int)(void):** Commits the changes that have been made until this point on the transaction, and returns the connection to the connection pool.
**rollbackTx:** Rollback any changes made in this transaction, and return the connection to the connection pool.

The following methods have been changed:
**Query(QueryRequest)()**: QueryRequest can now have a second type, which has subnodes 'txHandle' and 'query'. This is a way to expand this preexisting operation to support handling transactions.
**Update()():** Same as 'Query'
**Connect()():** General refactor to change from a long if-else block to a switch statement, as well as starting a ConnectionPool instead of a single connection.
**Close()():** Now closes the connection pool and resets the class.


javaServices/coreJavaServices/src/main/java/joliex/db/DatabaseService.java
```diff
@@ -115,148 +128,102 @@ public void connect( Value request )
...
 					connectionString += "?characterEncoding=" + encoding.get();
 				}
-				connection = DriverManager.getConnection(
-					connectionString,
-					username,
-					password );
+				connectionPool = _createDataSource();
 			}
 
-			if( connection == null ) {
```
Maje419, line 188: Using the connection pool, this check is unnecessary. The connection pool will fail to create and throw a SQLTransientConnectionException if it cannot validate any connections.

javaServices/coreJavaServices/src/main/java/joliex/db/DatabaseService.java
```diff
@@ -115,148 +128,102 @@ public void connect( Value request )
...
-					if( !connection.isValid( 0 ) ) {
-						connection = DriverManager.getConnection(
-							connectionString,
-							username,
-							password );
-					}
-				}
-			} catch( SQLException e ) {
-				throw createFaultException( e );
+			if( connectionPool.isClosed() ) {
```
Maje419, line 216: HikariCp will validate connections before handing them out, so it suffices to check whether the connection pool is closed.

javaServices/coreJavaServices/src/main/java/joliex/db/DatabaseService.java
```diff
@@ -523,24 +505,121 @@ static FaultException createFaultException( SQLException e ) {
...
-						request.getFirstChild( TEMPLATE_FIELD ) );
-				} else {
-					resultSetToValueVector( result, resultValue.getChildren( "row" ) );
-				}
-				result.close();
+		if( request.isString() ) {
+			try( Connection con = connectionPool.getConnection();
+				PreparedStatement stm = new NamedStatementParser( con, request.strValue(), request )
+					.getPreparedStatement() ) {
+				resultValue = _executeQuery( stm, request );
```
Maje419, line 514: Moved _executeQuery into its own method since it's being used two places

javaServices/coreJavaServices/src/main/java/joliex/db/DatabaseService.java
```diff
@@ -523,24 +505,121 @@ static FaultException createFaultException( SQLException e ) {
...
+			}
+		}
+	}
+
+	private HikariDataSource _createDataSource() {
+		HikariConfig config = new HikariConfig();
+
+		config.setUsername( username );
+		config.setPassword( password );
+		config.setMaximumPoolSize( 6 );
```
Maje419, line 614: I am very uncertain what would be a good pool-size, so for now it is hard-coded to 6.

[Microsoft sql](https://learn.microsoft.com/en-us/dotnet/framework/data/adonet/sql-server-connection-pooling) sets a connection pool size of max 100.

Connections are by default automatically added to the pool if needed, and removed after having been idle for 10 minutes, so maybe increasing this number is not bad

packages/database.ol
```diff
@@ -33,6 +33,25 @@ type ConnectionInfo:void {
 	.checkConnection?:int // if true (> 0) check connection before each DB command (default: false (0))
 	.toLowerCase?: bool // lowercase attribute names
 	.toUpperCase?: bool // uppercase attribute names
+	.connectionPoolConfig?: ConnectionPoolConfig
+}
+
+// See https://github.com/brettwooldridge/HikariCP?tab=readme-ov-file#frequently-used for descriptions
+type ConnectionPoolConfig: void {
```
Maje419, line 40: I left out the following configuration options on purpose:

- autoCommit?: bool	<-- The way the DatabaseService is currently implemented, providing this would simply mean that 'update' and 'query' breaks outside using transactions.
- metricRegistry	<-- Needs to be passed an object
- healthCheckRegistry <-- Needs to be passed an object
- allowPoolSuspension?: bool	<-- The logic for pool suspension is not implemented in the DatabaseService, so we're not allowing this for now
- registerMbeans <-- MBeans requires the user to modify the DatabaseService code, which is not intended
- driverClassName?: string 	<-- This is entered in the ConnectionInfo already, so just resolve it from there
- leakDetectionThreshold?: int <-- Logging is currently not enabled for DatabaseService, so this would be pointless
- dataSource?: DataSource <-- Requires to send a object of type DataSource
- threadFactory?: ThreadFactory <-- Requires user to provide an object of type ThreadFactory
- scheduledExecutor?: ScheduledExecutor <-- Requires user to provide object o f type ScheduledExecutor
- leakDetectionThreshold: int <- Since we are not logginng any information, this is pointless

javaServices/coreJavaServices/src/main/java/joliex/db/DatabaseService.java
fmontesi: Should probably be a configuration parameter.

mwallnoefer:
Hi @Maje419 this looks pretty good!
Have you thought to fix also issue #388 within your contribution? I guess it should not be that hard to address this too.
Further, have you tried to extend the database unit tests under `tests`? Just to keep this connection pool support working also in future.

Maje419:
> Hi @Maje419 this looks pretty good! Have you thought to fix also issue #388 within your contribution? I guess it should not be that hard to address this too. Further, have you tried to extend the database unit tests under `tests`? Just to keep this connection pool support working also in the future.

Thanks for taking a look, and apologies for the late response!

This PR does not include a fix #388. I'd love to implement a fix, but It'll take me a bit of time, as I then need to look into how connection strings are built, as well as write tests to fit. I'd prefer if I could make this fix in a separate fix at a later time?

In regards to tests, I already have several unit tests running in Jot, so it should not be difficult to port those over. I don't see any tests for the Database Service already present in the `test` folder, are they located somewhere else?

mwallnoefer (commented):
Very good, just the indentation does not always look consistent. Have you mixed spaces with tabs?

PR #494, opened 2024-04-10
Jolie2Java2

Jonaythaz:
This is a preview of the new Jolie2Java tool, with the details of how the tool now works being available in the **README.md** file of the **tools/jolie2java** folder.

Outside of those made within the **jolie2java** module, the changes are as follows:

- Updated the **Dockerfile** of the **.devcontainer** folder to use the newest version of its image (since Jolie2Java now requires Java 21)
- Updated the **jolie2java** executables in the **launchers** folder to include the dependencies of the new Jolie2Java tool
- Added files used by the classes generated by Jolie2Java to the **jolie** module (In the jolie.runtime.embedding.java package)
- Updated the **jolie** module to use Java21
- Added a **newJavaServices** module inside the **javaServices** folder, containing examples of using the new Jolie2Java tool.
- Updated the main **pom.xml** file to include **newJavaServices** as a submodule. 

Note that the last 2 changes were only made in order to provide an example of the new tool in action, and should be undone before an actual pull request.


mwallnoefer:
I guess that this requires a rebase which drops my commits from `master`, otherwise the diff is not readable at all.

Jonaythaz:
> I guess that this requires a rebase which drops my commits from `master`, otherwise the diff is not readable at all.

Yes that would probably be a good idea, I will do that before making the actual pull request

Jonaythaz:
Updated description of the changes (outside the **jolie2java** module):

- Updated the **jolie** module to use Java21
- Added files used by the classes generated by Jolie2Java to the **jolie** module (In the **jolie.runtime.embedding.java** package)
- Updated the **Dockerfile**s of the **.devcontainer** and **docker** folders to use the Java 21 version of their images
- Updated the **jolie2java** executables in the **launchers** folder to include the dependencies of the new Jolie2Java tool

mwallnoefer:
@Jonaythaz Regarding to Java versioning please let me ask the following question: is there any particular requirement for Java 21? I am asking this since with my latest cleanups I have ported the whole Jolie codebase to Java 11 compatibility, which is a minimum standard also for a bit older GNU/Linux distros like Ubuntu 22.04 LTS.
In general I would always try to keep the version requirement as low as possible, for instance if you are using features from Java 16 then ask for that one and not 21.   

@fmontesi Is the merge branch [jolie:feature/new-jolie2java](https://github.com/jolie/jolie/tree/feature/new-jolie2java) correct? So it should then not land directly into _master_?

Jonaythaz:
> @Jonaythaz Regarding to Java versioning please let me ask the following question: is there any particular requirement for Java 21? I am asking this since with my latest cleanups I have ported the whole Jolie codebase to Java 11 compatibility, which is a minimum standard also for a bit older GNU/Linux distros like Ubuntu 22.04 LTS. In general I would always try to keep the version requirement as low as possible, for instance if you are using features from Java 16 then ask for that one and not 21.
> 
> @fmontesi Is the merge branch [jolie:feature/new-jolie2java](https://github.com/jolie/jolie/tree/feature/new-jolie2java) correct? So it should then not land directly into _master_?

Java 21 is "required" because enhanced switch expressions are used throughout the generation process, and because it is the intended way to access the data of choice types in Java. The reason for merging into this branch and not master is because the new jolie2java is planned to be introduced together with the release of Jolie 1.13, where I believe the intention is to update Jolie in general to be using Java 21.

fmontesi:
This is looking pretty good now I think.

@Jonaythaz are there any outstanding issues or can we review it as is now? Also, could you make it a pull request towards `master` instead of another branch?

Jonaythaz:
> This is looking pretty good now I think.
> 
> @Jonaythaz are there any outstanding issues or can we review it as is now? Also, could you make it a pull request towards `master` instead of another branch?

I don't believe there are any issues that I'm aware of, so it should be fine to review it. The pull request should be towards `master` now.

PR #496, opened 2024-04-10
Fix/refactoring

mwallnoefer:


jolie-cli/src/main/java/jolie/cli/CommandLineParser.java
```diff
@@ -339,12 +339,24 @@ public CommandLineParser( String[] args, ClassLoader parentClassLoader, Argument
 			} else if( "--connlimit".equals( argsList.get( i ) ) ) {
 				optionsList.add( argsList.get( i ) );
 				i++;
-				cLimit = Integer.parseInt( argsList.get( i ) );
+				try {
+					cLimit = Integer.parseInt( argsList.get( i ) );
+				} catch( Exception e ) {
+					System.out
```
fmontesi, line 345: Mmh we should probably throw a CommandLineException, for consistency with the other error messages.

jolie-cli/src/main/java/jolie/cli/CommandLineParser.java
```diff
@@ -339,12 +339,24 @@ public CommandLineParser( String[] args, ClassLoader parentClassLoader, Argument
...
+				}
 				optionsList.add( argsList.get( i ) );
 			} else if( "--responseTimeout".equals( argsList.get( i ) ) ) {
 				optionsList.add( argsList.get( i ) );
 				i++;
-				rTimeout = Long.parseLong( argsList.get( i ) );
+				try {
+					rTimeout = Long.parseLong( argsList.get( i ) );
+				} catch( Exception e ) {
+					System.out
```
fmontesi, line 356: Idem, command line exception.

extensions/jsonrpc/src/main/java/jolie/net/JsonRpcProtocol.java
```diff
@@ -381,14 +395,14 @@ private CommMessage recv_createCommMessage( int messageSize, byte[] messageConte
 					"/", value.getFirstChild( "params" ), null );
 			} else if( value.hasChildren( "error" ) ) {
 				String operationName = jsonRpcOpMap.get( jsonRpcId );
-				return new CommMessage( Long.parseLong( jsonRpcId ), operationName, "/", null,
+				return new CommMessage( jsonRpcId.hashCode(), operationName, "/", null,
```
fmontesi, line 398: What's the reason for this change?

mwallnoefer: We should not silently ignore `NumberFormatExceptions`, but I have fixed this in the new version of the commit.

jolie-cli/src/main/java/jolie/cli/CommandLineParser.java
mwallnoefer: Here I based myself to the parsing of `cellId`. https://github.com/jolie/jolie/blob/c714390e8aba275c2cb677320a1d5219bdd0704e/jolie-cli/src/main/java/jolie/cli/CommandLineParser.java#L440.

But of course we may change the behaviour also there.

mwallnoefer: Okay I have fixed the two calls and push the refreshed commits.

fmontesi: Super, thanks! :)

PR #499, opened 2024-04-19
Fix/basictype

mwallnoefer:
Here I have found a better way to exploit the generics in the `BasicType` class' `fromBasicTypeDefinition()` method: commit 808d6ad2e090. But we need to introduce another unchecked cast in this new `checkRefinement()` helper method.

### DIFF
--- a/jolie/src/main/java/jolie/runtime/typing/BasicType.java
+++ b/jolie/src/main/java/jolie/runtime/typing/BasicType.java
@@ -23,7 +23,7 @@ import jolie.lang.NativeType;
 import jolie.lang.parse.ast.types.BasicTypeDefinition;
 import jolie.lang.parse.ast.types.refinements.BasicTypeRefinement;
 import jolie.runtime.Value;
-
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
@@ -32,8 +32,8 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
-public class BasicType< T extends Object > {
-	private static final Map< NativeType, BasicType< ? extends Object > > PURE_BASIC_TYPES = new HashMap<>();
+public class BasicType< T > {
+	private static final Map< NativeType, BasicType< ? > > PURE_BASIC_TYPES = new HashMap<>();
 	private static final Predicate< Value > ANY_PREDICATE = v -> true;
 	private static final Predicate< Value > INT_PREDICATE = Value::isInt;
 	private static final Predicate< Value > LONG_PREDICATE = v -> v.isLong() || v.isInt();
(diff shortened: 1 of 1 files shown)


jolie/src/main/java/jolie/runtime/typing/BasicType.java
```diff
@@ -81,6 +81,15 @@ public NativeType nativeType() {
 		return nativeType;
 	}
 
+	@SuppressWarnings( "unchecked" )
+	private static < T > void checkRefinement( Object nativeValue, BasicTypeRefinement< T > refinement,
```
fmontesi, line 85: Why not make the type of `nativeValue` a `T`?

mwallnoefer: This did not work by means the `for` loop:
```
			for( BasicTypeRefinement< ? > refinement : refinements )
				checkRefinement( nativeValue, refinement, pathSupplier );
```

On one place you need to weaken the generics, otherwise the underlying issue may not be solved (to have this combined data-type with two kinds of generics: the basic type with the refinements of some other type).

mwallnoefer: E.g. you could do this, but it is not much better:
```
	@SuppressWarnings( "unchecked" )
	private static < T,U > void checkRefinement( U nativeValue, BasicTypeRefinement< T > refinement,
		Supplier< String > pathSupplier ) throws TypeCheckingException {
		if( !refinement.checkValue( ( T ) nativeValue ) ) {
...
```

mwallnoefer: Any thoughts?

fmontesi: What if you add the constraints 'U extends T' in the generic parameter? Does that help a bit maybe? (This might require updating the signature of the refinements as well)

mwallnoefer: The problem here is that you are not able to solve it completely by not making also `BasicTypeDefinition` generic for the refinements which I wanted to avoid. And also then I guess that the discrepancy remains:
```
public static BasicType< ? > fromBasicTypeDefinition( BasicTypeDefinition basicTypeDefinition ) {
	return fromBasicTypeDefinitionInternal( PURE_BASIC_TYPES.get( basicTypeDefinition.nativeType() ),
			basicTypeDefinition.refinements() );
}
```
We are not able to concile the `List< BasicTypeRefinement< ? super T > >` for a certain basic type which we grasp from the map `PURE_BASIC_TYPES` with some additional `List< BasicTypeRefinement< ? > >` from the `refinements()` method.

So I have opted for creating this `fromBasicTypeDefinitionRefinements()` method which allocates a new list and passes the entries one by one. I could have made it also with a whole-list cast `(List<BasicTypeRefinement< ? super T >>)((List) ...)` to avoid the additional allocation but it didn't look like a very clean approach.

PR #505, opened 2024-05-10
Fix/various

mwallnoefer:


mwallnoefer:
Just one more question, @fmontesi (file _HttpProtocol.java_):

```
	private void recv_readJolieHeaders( HttpMessage message, HttpUtils.DecodedMessage decodedMessage ) {
		String messageId = message.getProperty( HttpUtils.Headers.JOLIE_MESSAGE_ID );
		if( messageId != null ) {
			try {
				decodedMessage.id = Long.parseLong( messageId );
			} catch( NumberFormatException e ) {
			}
		}

		if( message.isResponse() ) {
			String operationName = message.getProperty( HttpUtils.Headers.JOLIE_OPERATION );
			if( operationName != null ) {
				decodedMessage.operationName = operationName;
			}
		}
	}
```

If we are provided with the `id` but **NOT** the operation name is it not possible to deduce it from the internal hashtable?

fmontesi:
> Just one more question, @fmontesi (file _HttpProtocol.java_):
> 
> ```
> 	private void recv_readJolieHeaders( HttpMessage message, HttpUtils.DecodedMessage decodedMessage ) {
> 		String messageId = message.getProperty( HttpUtils.Headers.JOLIE_MESSAGE_ID );
> 		if( messageId != null ) {
> 			try {
> 				decodedMessage.id = Long.parseLong( messageId );
> 			} catch( NumberFormatException e ) {
> 			}
> 		}
> 
> 		if( message.isResponse() ) {
> 			String operationName = message.getProperty( HttpUtils.Headers.JOLIE_OPERATION );
> 			if( operationName != null ) {
> 				decodedMessage.operationName = operationName;
> 			}
> 		}
> 	}
> ```
> 
> If we are provided with the `id` but **NOT** the operation name is it not possible to deduce it from the internal hashtable?

No (there's no internal hashtable anymore). However, it should never be the case unless we're communicating with an old version of Jolie. (Which should now be considered deprecated for .concurrent as you've pointed out in the docs.)

fmontesi:
It all looks good, I just have a doubt about the ValueImpl-RootValueImpl refactoring. If I understand correctly, now we would pay a slight performance hit in accessing variables in the state of processes (`State` uses `RootValueImpl`) because the `children` field is now an indirect ref to the hash map. I wonder how big of a hit it actually is...

Maybe if you remove that part we could merge all the rest in the meantime?

I'm also trying to figure out if I can make `Lazy` (from `libjolie`) offer a static method to obtain a thread-safe `Lazy`, which might be helpful here.

mwallnoefer:
I have changed the commit so that `RootValueImpl` uses the `children` member without the `Atomic` wrapper, while `ValueImpl` continues to do so.

fmontesi:
Nice! However, now all values have the children map, always initialised, whereas the current implementation would initialise it lazily (only if necessary). Most values do not need that map so I'd imagine this would increase memory usage and decrease performance. Whereas for root values it makes sense to initialise it immediately and not have it indirect because root values make no sense without children (since Jolie offers no syntax to access the native value in the root of the value used in `State`).

fmontesi:
Maybe a solution lies in making a superclass for both ValueImpl and RootValueImpl, which leaves children() abstract or something. Maybe even `Value` should just be an interface in an ideal world...

mwallnoefer:
> Maybe a solution lies in making a superclass for both ValueImpl and RootValueImpl, which leaves children() abstract or something. Maybe even `Value` should just be an interface in an ideal world...

Ah ok, but then this comment over `RootValueImpl` was very misleading, otherwise I wouldn't have started this attempt at all:
```
/** TODO: remove code duplication from ValueImpl */
```

fmontesi:
> > Maybe a solution lies in making a superclass for both ValueImpl and RootValueImpl, which leaves children() abstract or something. Maybe even `Value` should just be an interface in an ideal world...
> 
> Ah ok, but then this comment over `RootValueImpl` was very misleading, otherwise I wouldn't have started this attempt at all:
> ```
> /** TODO: remove code duplication from ValueImpl */
> ```
> 
> Okay, I have excluded this commit so you may merge the rest. I will try to repropose it in a different way.

Well, I *would* like to get rid of that huge copy-paste between the two classes. But it is definitely subtle.

Btw I've tried to make jolie.util.Lazy efficient and thread-safe, it's in master.

fmontesi (approved):

fmontesi:
Rest is awesome, thanks!

Let me know if I can help with the refactoring of value if you have any doubts.

PR #521, opened 2024-09-19
Interpreter.init waits until service is initialised

kicito:
Introduce a Future to the InitDefinitionProcess, which informs the service initialization status.
Move the execution of input port processes to execute first during the initialization.

### DIFF
--- a/jolie/src/main/java/jolie/Interpreter.java
+++ b/jolie/src/main/java/jolie/Interpreter.java
@@ -82,6 +82,7 @@ import jolie.net.CommMessage;
 import jolie.net.SessionMessage;
 import jolie.net.ports.OutputPort;
 import jolie.process.DefinitionProcess;
+import jolie.process.InitDefinitionProcess;
 import jolie.process.InputOperationProcess;
 import jolie.process.SequentialProcess;
 import jolie.runtime.FaultException;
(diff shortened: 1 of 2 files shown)


jolie/src/main/java/jolie/process/InitDefinitionProcess.java
```diff
@@ -91,11 +123,14 @@ public void run()
...
 				super.run();
 			} catch( ExitingException e ) {
+				this.executionDone.complete( e );
 			}
 		} catch( EmbeddedServiceLoadingException e ) {
 			interpreter.logSevere( e );
+			this.executionDone.complete( e );
 		}
 	}
 }
```
fmontesi, line 136: The future might never be completed in the case of unchecked exceptions.

jolie/src/main/java/jolie/Interpreter.java
```diff
@@ -1004,7 +1014,9 @@ private void init( State initState )
 		} else {
 			sessionStarters = Collections.unmodifiableMap( sessionStarters );
 			try {
-				initExecutionThread = new InitSessionThread( this, getDefinition( "init" ), initState );
+				InitDefinitionProcess initProcess = (InitDefinitionProcess) getDefinition( "init" );
+				var future = initProcess.getFuture();
+				initExecutionThread = new InitSessionThread( this, initProcess, initState );
```
fmontesi, line 1020: Why don't you just make the future a parameter of the constructor of `InitSessionThread`? It should give a much simpler implementation.

fmontesi (changes requested):

jolie/src/main/java/jolie/process/InitDefinitionProcess.java
```diff
@@ -31,15 +33,54 @@
...
 
 public class InitDefinitionProcess extends DefinitionProcess {
+
+	/**
+	 * A `CompletableFuture` that holds the result of the execution.
+	 *
+	 * It is completed with either `null` if the initialization of the service is successful or an
+	 * `Exception` if an error occurs.
+	 */
+	private final CompletableFuture< Exception > executionDone;
```
fmontesi, line 43: Mmmh it's a bit weird to have this in a node of the interpretation tree. They are supposed to be immutable/as stateless as possible.

Maybe we could put the future in the `Interpreter` class?

PR #524, opened 2024-09-22
Min values for number ranges

Niels-Erik:
Added the possibility for '-*' in ranges of type refinements, and did some refactoring and commenting of peculiarities.


fmontesi:
@jolie/core Can I get a double check on using `-*` as syntax for 'minimum integer or double'? Please comment.

fmontesi:
Basically you can now write things like `ranges( [-*,10] )` to express an
integer from MIN_INT (which is negative) to 10. We cannot currently express
MIN_INT in ranges.

What I'm asking is: is this a good syntax for expressing MIN_INT?

klag:
But the idea is that, when there is '-*' there is no lower limit, right?

fmontesi:
Right (well, no limit on top of what the type supports, like int can't go
lower than MIN_INT)

PR #530, opened 2024-10-02
Servicenode implements DocumentedNode

kicito:
This PR allows documentation string to be stored in the ImportStatement and ServiceNode. The information will be used for the generated joliedoc in the future, and this also fixes error line on lsp when putting a joliedoc comments around those nodes e.g.

```
/**! test */
from console import Console    <--- error unexpected test
...
```


fmontesi:
I understand why we might wanna comment a `service` block, but why would we want to document an `import`? Where would we display that in the generated doc?

kicito:
I thought having a side-by-side code/comments HTML document could be useful for going through the code for learning purposes. Thinking again we may not need it for now, as we might be able to do it externally. PR is updated. :)

fmontesi:
Looks like this requires fixing inspector, the tests do not pass. :)

PR #531, opened 2024-10-11
fix for showing line number correctly when using --trace

Niels-Erik:
When running a file with --trace the line numbering was incorrectly printed, E.G. these should be 62 and 204:
/workspaces/jolie/ex1.ol:6297.     << SR
/workspaces/jolie/ex1.ol:20447.   :: ASSIGNMENT

This happens because there is no whitespace between the line number and actionCounter.

This commit fixes that, so the lines look like:
/workspaces/jolie/ex1.ol:62     97.     << SR
/workspaces/jolie/ex1.ol:204   47.     :: ASSIGNMENT


fmontesi:
Nice, thank you!

PR #536, opened 2024-11-05
Meta jolie2java code generation

kicito:


packages/meta.ol
```diff
@@ -42,8 +41,7 @@ type IntRefinement {
 }
 
 /// A long basic type.
-type LongBasicType {
-	long
+type LongBasicType: long {
```
fmontesi, line 44: This and the other *BasicType are not mistakes. I actually need a subnode called 'long' or something like that. Maybe we can use an annotation to generate something with a different name in Java?

fmontesi (changes requested):

packages/meta.ol
kicito: Java reserves void, int, enum, long, and double, so we'd need different names.

kicito: would is... e.g. `isInt` make sense?

fmontesi: Yes. :-)

PR #537, opened 2024-11-07
Fix/mustache recursive templates

klag:


packages/mustache.ol
```diff
@@ -22,13 +22,17 @@ type RenderRequest {
 	template:string ///< The mustache template
 	data:undefined ///< The data for the template
 	dir?:string ///< The directory in which to look for other templates (for partials)
+	recursionLimit?: int ///< Maximum limit for recursive calls in partials. Default is 100
+	partialsRecursionLimit?: int ///< Maximum limit for recursive calls specific for partials. Dafult is 10
```
fmontesi, line 26: What is exactly the difference between `recursionLimit` and `partialsRecursionLimit`? Is there another way to do recursion than using partials?

PR #539, opened 2024-11-12
fix JSON converter finds correct subtype if subtypes in TypeChoice has same name

Niels-Erik:
added findSubType that takes a String and a Value, and finds the Type which also corresponds to the Value
valueToJsonString in JsUtils.java now uses the new method to find the correct subtype

This is to fix the issue reported in discord:
sundedrengen — 10/29/2024 2:02 AM
We are pretty sure we found a bug in the way Jolie generates json reponses in the following circumstance (tested on jolie commit 3b8c752):
```
//Doesn't happen if type is 
//type Response: B | A
type Response: A | B

type A {
    canbeanything*: undefined
    test: string
}

type B {
    test?: A
}

interface Iface {
    RequestResponse: hello(void)(Response)
}

service BugHunter {
    execution: concurrent
    inputPort ip {
        location: "socket://localhost:12345"
        protocol: http {format = "json"}
        interfaces: Iface
    }
    main {
        [hello()(Response) {
            Response << {
                test << {
                    canbeanything[0] = "This should be wrapped in an array"
                    test = ""
                    //Not wrapped because something with the same name (in this case "test") exists in both of the types (we think).
                    /*
                    curl localhost:12345/hello
                    {"test":{"test":"","canbeanything":"This should be wrapped in an array"}}
                    */
                }
            }
         }]
    }
}
```

Not sure if it is done in the correct way.

### DIFF
--- a/jolie/src/main/java/jolie/runtime/typing/Type.java
+++ b/jolie/src/main/java/jolie/runtime/typing/Type.java
@@ -60,7 +60,7 @@ class TypeImpl extends Type {
 		}
 
 		try {
-			checkSubType.check( value ); //checks if the value is of the the given type (I think)
+			checkSubType.check( value );
 			return checkSubType;
 		} catch( TypeCheckingException e ) {
 			return null;
--- a/jolie/src/main/java/jolie/runtime/typing/Type.java
+++ b/jolie/src/main/java/jolie/runtime/typing/Type.java
@@ -21,16 +21,15 @@
 
 package jolie.runtime.typing;
 
+import java.util.HashMap;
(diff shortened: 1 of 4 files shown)


fmontesi:
Agreed, a test would be nice. Otherwise the PR looks good! :)

Niels-Erik:
Thank you, I'll add a test.

Niels-Erik:
@fmontesi @klag
I've managed to create a test testing for the specific scenario.
I've tested that it fails on master and we can see from the check passing that it succeeds in this branch.

PR #546, opened 2024-12-17
Switch to using a Parsing Context based on LSP Location

kas0079:
This PR introduces a new Parsing Context based on LSP, replacing the old URIParsingContext.

I kept the methods of the old URIParsingContext in the new ParsingContext.

Besides changing other classes to construct the new ParsingContext, I had to set `tools/jolie2wsdl` to compile using Java 21, since the new ParsingContext (which it uses) is a record class.


kas0079:
This PR is more complicated than it needs to be,  #563 accomplishes the same in a much simpler way. We can close this if that one is merged.

PR #548, opened 2024-12-22
DatabaseService fixes

mwallnoefer:


fmontesi:
I'd like to merge this, but there's an issue in the test: it re-downloads the derby jar every time it's run. This will slow down local testing too much. Can we make it a dependency in `test/pom.xml` perhaps? @klag @mwallnoefer @kicito

mwallnoefer:
Ah, Derby should get included in [_test/library/private/DB-test/pom.xml_](https://github.com/jolie/jolie/blob/master/test/library/private/DB-test/pom.xml).

And then all Derby DB paths have to be changed into `file:library/private/DB-test/testDatabase/` ([_test/library/database.ol_](https://github.com/jolie/jolie/blob/master/test/library/database.ol)).

klag:
Could we merge this? @fmontesi

fmontesi:
Great work everyone, thanks. :)

Issue #550, opened 2024-12-23
wrong import not detected


klag:
in this branch

https://github.com/jolie/jolie/tree/bug/wrong-import-not-detected

I added a  test called `import2.ol` where there is a wrong import that is not detected because the resource is defined in the same module.

kicito:
The PR #551 should fix this issue. It causes the test for metarender-code to fail, is it expected?

klag:
Do you mean this? 
https://github.com/jolie/jolie/blob/bug/wrong-import-not-detected/test/library/metarender-code.ol

It should not raise an error.
Which one it raises?

kicito:
It complained about missing the NativeCodeRenderInterface module. I see that a new commit on master has removed that line. I'll try to rebase and update the PR.

mwallnoefer:
Okay, PR #551 should fix the issue, so let's close this.

PR #551, opened 2025-01-05
Fix ModuleCrawler.crawlModule's logic

kicito:
Get the list of modules to crawl based on ImportStatement rather than from SymbolTable. This fixes the unexpected behavior if the imported symbol is overridden, which results in an ignored import statement, as mentioned in #550.


mwallnoefer:
The `metarender-code` test is failing since it expects the `NativeCodeRenderInterface` to be located in an `.ol` file, not in an `.iol` one. I guess we need to check for both.

PR #561, opened 2025-01-21
refactored joliemock in new syntax and made it available in packages

klag:
Refactored joliemock. The generated code does not have a perfect indentation but it works.

### DIFF
--- a/test/services/private/testservice.ol
+++ b/test/services/private/testservice.ol
@@ -92,6 +92,7 @@ interface DemoInterface {
 execution{ concurrent }
 inputPort DEMO {
   Location: "local"
+  Protocol: sodep
   Interfaces: DemoInterface
 }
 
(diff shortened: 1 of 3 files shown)


test/services/private/testservice.ol
mwallnoefer, line 1: I guess this change is not necessary. Location _"local"_ should not require any specific protocol. Therefore I have sorted out all those superfluous directives with a big refactoring commit in spring last year.

klag: it is because metarender produces an input port like this one:

inputPort X {
 location: "local"
 protocol: 
 interfaces: MyInterface 
}

So maybe, we need to fix metarender code that in case there is only location "local" and no protocol, protocol can be skipped

mwallnoefer: Ah ok, I have got it. Sorry for the noise but I had no idea that this code was auto-generated.

PR #581, opened 2025-02-07
Fix the ParsingContext of Services via a new wrapper for generating ParsingContexts

kas0079:
Adds a wrapper function `parseInternals` to make it easier to generate correct `ParsingContexts` when parsing.

Also refactors `parseService` to generate its `ParsingContext` using that method, fixing its previously incorrect `parsingContext`


mwallnoefer:
@kas0079 Should PR #605 replace this one?

kas0079:
> @kas0079 Should PR #605 replace this one?

Yes, while #605 is technically wraps a different method, there is a lot of overlap and it fixes some deeper problems with the implementation of the wrapper that this PR doesn't.
So maybe it is better to close this one, then I can remake it in a simpler form once #605 is merged.

PR #583, opened 2025-02-18
broken extender property on embedded services

mwallnoefer:


mwallnoefer:
@Niels-Erik I have managed to fix the type-checker issue with d1a971e838ea23142bf0a56a1e5c55fd63aa9087, but now the two test files seem to get in conflict by using the same channels?

Niels-Erik:
So if I understand correctly, we're not allowed to use `embed S as s ` with services that uses extenders anymore?
If this is the case, then we only need the one test, doesn't matter which.

I think it would be nice to create an enhancement issue to allow this in the future maybe?

What are peoples thoughts on this?

mwallnoefer:
@Niels-Erik Exactly, it does not work by means of the type checker (see my commit message).

So if I do not hear other opinions I propose the removal of `extender2.ol` from the patch.

kicito:
I've commit code that makes it support for embedding a service that exposed inputport with aggregation. In order to solve missing extender information when creating a binding Outputport (happens when running`embed .. as ..`), the OutputPortInfo now carries the InterfaceExtender and apply it's type during OOITBuilder phase. Thank you @Niels for the earlier commit that helps me pinpoint the problem easily. @fmontesi could you have a look?

libjolie/src/main/java/jolie/lang/parse/ast/PortInfo.java
mwallnoefer, line 1: Did interface duplicates happen also before or is this due to this patch? Could it make sense to convert the list into an ordered set?

kicito: Actually, after tidying the code and inspecting the content, the duplication disappeared. I removed the check

fmontesi:
@kicito I have some conflicts here, can you check?

kicito:
Which file has a conflict? I merged the master to this branch and did not get any.

fmontesi:
My bad, apparently the GitHub web UI was just telling me about a problem with rebasing. I squashed and merged instead.

Issue #582, opened 2025-02-12
URIParsingContext are usually wrong in some way


Niels-Erik:
In developing for #514 I have found that the URIParsingContexts provide wrong line numbers either in `startLine` method or the `enclosingCodeWithLineNumbers` method.

I have 2 examples, where the normal fault messege is printed followed by a line with `sourceName()` concatenated with `startLine()`, with the last line being the output of `enclosingCodeWithLineNumbers()`:

The first is in the called service where the fault comes from `throw( error, "an error" )` on line 25 according to VS code.
Here we see the `enclosingCodeWithLineNumbers()` method gives wrong line number and data:
```
Feb 12, 2025 5:59:48 PM thrower.ol
INFO: [throwembed2.ol -> thrower.ol -> bug] Thrown unhandled fault: error
Content (if any):string = "an error"
/workspaces/jolie/test/../faultlinesManualTests/thrower.ol:25
        26:            res 

```
The second is the caller which calls the called service with `add@test(1)(c)` on line 15.
Here we see the `startLine()` method giving the wrong line number:
```
Feb 12, 2025 5:59:48 PM throwembed2.ol
INFO: [throwembed2.ol] Thrown unhandled fault: error
Content (if any):string = "an error"
/workspaces/jolie/test/../faultlinesManualTests/throwembed2.ol:14
        15:        add@test(
```

I do not know if #581 fixes these cases.

mwallnoefer:
I guess also this should have been fixed? Otherwise please re-open.

Niels-Erik:
> I guess also this should have been fixed? Otherwise please re-open.

I was told some other people was working on things similar to this, so it is not fixed in #584, sadly I do not have permission to reopen the issue.

Issue #514, opened 2024-06-09
Provide line numbers with faults


fmontesi:
Debugging faults that reach the top level and are then logged by the Jolie interpreter gives errors messages like the next one.

```
Jun 09, 2024 12:38:19 PM main.ol
INFO: [main.ol] Thrown unhandled fault: MyFault
Content (if any):string = "My precious fault"
```

It would be helpful if this message contained also the line number in the source code, as one gets in Java with stack traces enabled. This way, a developer would know where to start their review of the code.

Niels-Erik:
I've looked into adding lin numbers to faults, I do have something small working, and I would like some input on how it should look.
Depending on where the fault is originated from it can look different:
```
Feb 06, 2025 1:03:04 PM throw.ol
INFO: [throw.ol] Thrown unhandled fault: error
Content (if any):string = "an error"
```
```
Feb 06, 2025 1:17:40 PM thrower.ol
INFO: [throwembed.ol -> thrower.ol -> bug] Thrown unhandled fault: error
Content (if any):string = "an error"
```
I would think that adding the line number inside the brackets with the names would look nice:
```
Feb 06, 2025 1:03:04 PM throw.ol
INFO: [throw.ol:14] Thrown unhandled fault: error
Content (if any):string = "an error"
```
```
Feb 06, 2025 1:17:40 PM thrower.ol
INFO: [throwEmbeder.ol:8 -> thrower.ol:16 -> serviceName] Thrown unhandled fault: error
Content (if any):string = "an error"
```
Or I could add new line like:
```
Fault happened through:
/workspaces/jolie/test/../thrower.ol:16
/workspaces/jolie/test/../throwembed.ol:8
```
Which would give space for the path too.

I might just start doing the second method as it would be easier to start with.

fmontesi:
```
Feb 06, 2025 1:17:40 PM thrower.ol
INFO: [throwEmbeder.ol:8 -> thrower.ol:16 -> serviceName] Thrown unhandled fault: error
Content (if any):string = "an error"
```
What's `throwEmbeder.ol:8` here? The line where thrower.ol was embedded? If so it might be a bit confusing, since instead `thrower.ol:16` is the line where the fault was generated instead. (So numbers have different meanings.)

I'd be happy with just having the last line number and the filename. I don't know where to put it exactly.. mmmh... maybe a new line like this?
```
Feb 06, 2025 1:17:40 PM thrower.ol
[...]

PR #584, opened 2025-02-20
Add location data / line numbers to faults

Niels-Erik:
Pull request for adding location data / line number to faults (see #514)

I've combed over every place that creates a fault and all the places that I could find that forwards them.

I've tested the following manually and seen that they report line numbers(nearly correctly see #582):

- throwing faults
- causing faults from arithmetic (divide by zero and ++ operator on non numbers)
- faults coming from calls (examples call@service()() )
- faults forwarded by forward statement
- faults send over Local and not (used sodep)
- TypeMismatch that are thrown as fault have context (some TypeMismatch arning are logged not using faults)

There exists some exception that can come up in ScopeProcess and OutputPort which I've not found ways to test.

### DIFF
--- a/jolie/src/main/java/jolie/Interpreter.java
+++ b/jolie/src/main/java/jolie/Interpreter.java
@@ -617,10 +617,12 @@ public class Interpreter {
 		StringWriter writer = new StringWriter();
 		try {
 			new ValuePrettyPrinter( fault.value(), writer,
-				"Thrown unhandled fault: " + fault.faultName() + "\nSource location: " + fault.getContextString() + "\nContent (if any)" ).run();
+				"Thrown unhandled fault: " + fault.faultName() + "\nSource location: " + fault.getContextString()
+					+ "\nContent (if any)" ).run();
 			logInfo( writer.toString() );
 		} catch( IOException e ) {
-			logInfo( "Thrown unhandled fault: " + fault.faultName() + "\nSource location: " + fault.getContextString() );
+			logInfo(
+				"Thrown unhandled fault: " + fault.faultName() + "\nSource location: " + fault.getContextString() );
 		}
 	}
 
--- a/jolie/src/main/java/jolie/Interpreter.java
+++ b/jolie/src/main/java/jolie/Interpreter.java
@@ -617,10 +617,10 @@ public class Interpreter {
 		StringWriter writer = new StringWriter();
 		try {
 			new ValuePrettyPrinter( fault.value(), writer,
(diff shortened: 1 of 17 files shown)


fmontesi:
Nice work! Two questions:
- Why `addContext`, when it seems like a setter method? Shouldn't it be called `setContext`?
- I'm wondering about that `CommMessage::faultClone` method. Can we avoid introducing it by calling `setContext` on the message's fault maybe?

I'm slightly concerned about making FaultException mutable with `setContext`. We could consider a `withContext` method that returns a new FaultException with the context set in, I guess, but could be overengineering. Any thoughts?

Niels-Erik:
> * Why `addContext`, when it seems like a setter method? Shouldn't it be called `setContext`?

Yes it should, it was a leftover from when I tested with a list of all contexts that the fault would could get, I'll change it.

> * I'm wondering about that `CommMessage::faultClone` method. Can we avoid introducing it by calling `setContext` on the message's fault maybe?

Sadly not, this is done to avoid an embedded service show the same fault line as the embedder. Setting the context to null in fault will still cause the same issue. The issue being that the embedded and the embedder points to the same fault.
I did think about using a clone of the fault when making the `CommMessage`, but since it is only a problem for local communication I opted for the `CommMessage::faultClone` which is only used one specific place.

> I'm slightly concerned about making FaultException mutable with `setContext`. We could consider a `withContext` method that returns a new FaultException with the context set in, I guess, but could be overengineering. Any thoughts?

I feel this would be overengineering, as I see no reason to create a new object when we want to add the context, it could be useful if we opted to handle the local communication issue outside of `LocalCommChannel`, but again I think this would break some rules in SOLID.

Niels-Erik:
FaultException are now immutable again by using `withContext`, this corrrects everything mentioned(https://github.com/jolie/jolie/pull/584#issuecomment-2680929679).

However there is now more overhead, as FaultExceptions are created just to create a nearly identical FaultException.
Creating more constructors would fix the overhead, but introduce twice the amount of constructors, should this be done?

fmontesi:
How many constructors are we talking about? If it's like adding 2 to 4 constructors, I'd do it.

Niels-Erik:
> How many constructors are we talking about? If it's like adding 2 to 4 constructors, I'd do it.

Short of it 6 new constructors.

There exists 6 which all potentially would be duplicated, and from my quick look through it was 5 or 6 needed, and then nearly all `withContent` methods could be removed.

PR #587, opened 2025-02-24
CommCore PollingThread - better use yield() for activating the next thread

mwallnoefer:


fmontesi:
I agree this will be faster, but won't it increase energy consumption by a fair bit since it'll enter the loop again very soon? `yield` tells the scheduler to reschedule the thread asap. Any thoughts?

mwallnoefer:
But we have `wait()` calls for getting into the WAITING state when idle, so I don't know if the overhead is really that big.

fmontesi:
You're right, we do. Let's go with this change then! Thanks.

Issue #508, opened 2024-05-25
JSON pretty printing


mwallnoefer:
Currently, Jolie does not provide any pretty-printing support for JSON, which could be very handy for writing human-readable output files like OpenAPI specifications. Unfortunately our current extension library `json-simple` does not provide any support so far, but maybe we replace it with a more functional component in the future.

https://stackoverflow.com/questions/4105795/pretty-print-json-in-java

mwallnoefer:
When we have the _jackson_ library added to Jolie we could implement JSON pretty printing by the means of the following code piece: https://stackoverflow.com/a/42527104/4320224

PR #593, opened 2025-02-27
Enhancement/json schema

klag:
Added json-schema package for converting jolie types into json types compatible with openapi 2 and 3 (it is possible to select the standard)


mwallnoefer:
@klag regarding test/json_utils.ol please consider my latest changes here: https://github.com/jolie/jolie/commit/36eb474b815128c101bb01eddeaa7c80751af1ce. I have introduced them to fix the `/` escape bug.

klag:
@mwallnoefer rebased

mwallnoefer:
@klag Btw, does your work address also the problem of pretty-printing JSONs (https://github.com/jolie/jolie/issues/508)?

klag:
Hi @mwallnoefer no. it just provides an api for converting jolie types into jsonschemas for both openapi 2.0 and 3.0

packages/json_utils.ol
```diff
@@ -29,6 +29,20 @@ type GetJsonValueRequest: any {
 }
 type GetJsonValueResponse: undefined
 
+type ValidateJsonRequest: void {
```
fmontesi, line 32: You don't need `: void`

packages/json_utils.ol
```diff
@@ -29,6 +29,20 @@ type GetJsonValueRequest: any {
...
+	json: string 
+	schema: string 
+}
+
+type ValidateJsonResponse: void {
```
fmontesi, line 37: You don't need `: void` (and similarly below).

In general I'd avoid writing `: void` any longer for void nodes that contain subnodes.

fmontesi (changes requested):

packages/json_utils.ol
klag: Done

klag:
@fmontesi ping

PR #596, opened 2025-03-02
added test for extender in embed without 'as' keyword

Niels-Erik:
In commit 18bfb4798235a3b93900ba284dd760fad53e9d76 I fixed partly the bug, as I only tested on a service that used embed without as.
So I made a test for this specifically(`extender2.ol`), which in a way is subset of what's being tested in `extender.ol`.

### DIFF
--- a/jolie/src/main/java/jolie/runtime/embedding/JolieServiceNodeLoader.java
+++ b/jolie/src/main/java/jolie/runtime/embedding/JolieServiceNodeLoader.java
@@ -26,6 +26,9 @@ import java.nio.file.Paths;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import jolie.Interpreter;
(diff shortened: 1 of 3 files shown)


mwallnoefer:
extender2.ol is ok, externer.ol unfortunately not:

```
extender.ol...                      ❌ failed (default handler).
SEVERE: [test.ol -> extender.ol -> Main] Request-Response process for test threw an undeclared fault for that operation (TypeMismatch), throwing TypeMismatch
test-template.ol...                 passed ✅
Mar 02, 2025 8:37:54 PM extender.ol
INFO: [test.ol -> extender.ol -> Main] Thrown unhandled fault: TypeMismatch
Content (if any):string = "Output message TypeMismatch (nothing@EmbedderService): Unexpected child node: #Message.token"
```

PR #597, opened 2025-03-03
Apply ModuleSource to interpreter

kicito:
This PR changes the following:

- Changes all interpreter constructors from Inputstream & URI to ModuleSource, each wtith a specific implementation of additional paths to include when scanning for lib.
- Adds new classes representing ModuleSource - InputstreamSource, and ProgramSource, to support different scenarios of interpreter creation. 
- Changes the default module lookup to cwd instead of the parent path of the parsing ModuleSource.
- Encapsulate the ModuleSource creation into a static `create` method.


mwallnoefer:
Very good @kicito that you are trying to eliminate more deprecated (Java 20+) `URL` constructors.

kicito:
Thank you, there are still some `new URL` around in cmdparser that I do not dare to touch yet. Windows seems to be not satisfied with it, though ;(

fmontesi:
Yeah, Windows is tricky when it comes to URLs. :\ Let me know how you fare.

kicito:
All right, after pingponging between linux and windows, the test now passes on all platforms. For cross-platform, I try to use URI for path-related Strings after the program passes cmdParser, which backfires when the program creates a new instance of olParser and cmdParser, specifically when parsing `include` or embed service before the modulesystem. A way forward for refactoring this is to decouple the cmdParser to a separate file, it does too much at the moment.

Issue #230, opened 2020-05-14
jolie2surface fails when there are services in jap aggregated into the input port


klag:
If I run jolie2surface for a service which embeds and aggregates another service that is contained in a jap, it fails

kicito:
the error is caused by lack of jap scheme handler in GetSurface class loader?
I just check and it is the case.

fmontesi:
It could be. Maybe we should just move the code that initialises the JAP Url handler in CommandLineParser?
Does anybody see counter-indications to this, @jolie/developers ?

mwallnoefer:
@kicito could this issue have been fixed meanwhile?

mwallnoefer:
I have tested the _mqtt client_server_ example and at least the interface and types get extracted out correctly from the jap (`Server` is a local output port in server.ol).
```
jolie2surface server.ol Server
```
But the following commands do not work (package: https://github.com/jolie/jolie-mosquitto/blob/master/JavaService/JolieMosquittoConnector/resources/packages/mqtt/mqtt.ol):
```
jolie2surface server.ol MQTT.ip
OR
jolie2surface server.ol ip
```

mwallnoefer:
@klag will PR #579 address also this issue?

klag:
Now jolie2surface is built on top of metajolie. I did not check if it works properly when jap files are considered. I think there are the same open problems we have so far with jap. Once they are solved it should work.

mwallnoefer:
> Now jolie2surface is built on top of metajolie. I did not check if it works properly when jap files are considered. I think there are the same open problems we have so far with jap. Once they are solved it should work.

You're right @klag, the bug persists also after applying your PR.

fmontesi:
@klag @mwallnoefer Can you test again? I've just merged #599

mwallnoefer:
> But the following commands do not work (package: https://github.com/jolie/jolie-mosquitto/blob/master/JavaService/JolieMosquittoConnector/resources/packages/mqtt/mqtt.ol):
> 
> ```
> jolie2surface server.ol MQTT.ip
> OR
> jolie2surface server.ol ip
> ```

Still not working with latest git:

```
$ jolie2surface server.ol MQTT
Input port MQTT not found
$ jolie2surface server.ol MQTT.ip
Input port MQTT.ip not found
$ jolie --version
Jolie 1.13.0-git  (C) 2006-2025 the Jolie developers
```

klag:
[...]

PR #599, opened 2025-03-04
Jap Module System

kicito:
This PR is the continuation of the applying ModuleSource PR. This PR unlocks the module system to be usable within jap file.


mwallnoefer:
Does it also address issue #230 ?

PR #600, opened 2025-03-06
fixed mit licenses

klag:


mwallnoefer:
@klag could we remove all these trailing white-spaces before merging? I am not a big friend of them, since the git tools show them in bold red. And after all it is just the matter of applying some simple regex.

fmontesi:
That would indeed be nice! I'm merging this in the meantime since it contributes to an important issue.

PR #604, opened 2025-03-26
added getHashMap api

klag:
very useful functionality. It should be implemented as a native statement into the language.
It could be nice to have a transformation from vectors to maps and viceversa

an example of usage can be found in the test

```
for( x = 0, x < 100, x++ ) {
    a.c.d.f[ x ] << {
        testString = "testString"
        mykey = "key" + x
        x = x
    }
}
getHashMap@Trees( {
    vector -> a.c.d.f 
    key = "mykey"
})( hashmap )

count = 0
foreach( y : hashmap ) {
    count++
    if ( y != hashmap.( y ).mykey ) {
        throw( TestFailed, "Key not correspond, expected " + y + " found " + hashmap.( y ).mykey )
    }
}

if ( count != 100 ) {
    throw( TestFailed, "expected 100 keys, found " + count )
}
```

### DIFF
--- a/packages/trees.ol
+++ b/packages/trees.ol
@@ -1,0 +1,55 @@
+/*
+ *   Copyright (C) 2025 by Claudio Guidi <cguidi@italianasoftware.com>          
+ *                                                                         
+ *   This program is free software; you can redistribute it and/or modify  
+ *   it under the terms of the GNU Library General Public License as       
+ *   published by the Free Software Foundation; either version 2 of the    
+ *   License, or (at your option) any later version.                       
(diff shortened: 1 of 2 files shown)


klag:
At the moment is quite limited because we should need Path variables which would allow to specify a path for the key and not only a sinfle field name:

Could we consider the idea to introduce path variables inside Jolie?

I think it is sufficient to have

```
myrootnode.("a.b.c.d") 
```

that is evaluated as 
```
myrootnode.a.b.c.d
```

@fmontesi

PR #605, opened 2025-03-27
Introduce parsingContext wrapper and refactor _parseExecutionInfo using it

kas0079:
This PR Introduces a wrapper `parseInternals` intended to make creating `parsingContexts` in the various parsing methods of `OLParser` easier and some prerequisites for it in the Scanner and Parser.

This PR is similar to #581, for which @fmontesi, suggested I start with a simpler parsing method like _parseExecutionInfo

Changes I made in `Scanner`:
 - I introduce a variant of `currColumn` called 'currOffset' which unlike `currColumn` does not increment multiple times on '\t' characters, the behavior of the former makes it hard use in with for example [LSP Range](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#range) which does not treat '\t' different from other characters. Additionally I made the Offset 0-indexed since I need in my use-case anyways.
 - I added 4 ints for tracking the start and end of `Tokens`
 ```java
	private int tokenStartLine;	// Line the last returned token started on.
	private int tokenStartOffset;	// Offset that the last returned token started on
	private int tokenEndLine;	// Line the last returned token ended on
	private int tokenEndOffset;	// Offset the last returned token ended on
```

In `AbstractParser` I added functionality to get the `Scanner` `tokenEndLine` and `tokenEndOffset` as they were one token ago, this is useful because the parsing methods in `OLParser` increment the Token stored in the Parser after parsing what they need, e.g.
```jolie
service Main {
    inputPort Foo {
        ...
    } /* <- after eating this } the Parser.token will be the following inputPort keyword, 
	so info about the previous token needs to be saved.
    */
    inputPort Bar {
        ...
    }
}
```

**What remains to be decided is:**
 - Since I introduced a distinction between `currColumn` and `currOffset` in the scanner, how do I best save the `currOffset` in the `ParsingContext`?
 I am considering adding new fields for them:
 ```java
 public class URIParsingContext {
	private final int startColumn;
	private final int endColumn;
	private final int startLine;
	private final int endLine;
	private final Optional<Integer> startOffset = Optional.empty();
	private final Optional<Integer> endOffset = Optional.empty();
}
```
and a new constructor that also sets them.
I want to use `Optional` because `ParsingContext` is used in many places that don't provide offsets yet, and I think it is more clear than using special int values like ´-1´ to signal it.
- I introduced 4 int tokenStart/End member variables to `Scanner`, I think it would be more elegant if those 4 are added to the Token class instead, but it would probably at least double the size of my PR considering there are 60 calls to the `Token` constructor, should I try to do it anyways?.
- I created a `class SaveLastN` (a FIFO structure that discards the oldest element on insertion if it is full.), I use it in the `AbstractParser` with `capacity = 2` to save the previous Token end as mentioned above. I am wondering if it is too niche a class to have it separately and if it should be a nested class of `AbstractParser` instead?


kas0079:
@kicito I did the changes you asked for, so it should be ready to merge right?

Issue #320, opened 2021-05-23
 jolie2openapi: "Schemes" tag is missing in the generated swagger json


scorsagg:
When I generate the swagger json using jolie2openapi and use it to open on swagger hub, I get default URL of https and not able to modify that. 

The "schemes": ["http"], is missing in the generated openapi definition.

Please provide an option to add this.

mwallnoefer:
@klag ping, what do you think? Since you were the original author of the tool, do you share the opinion that we should be adding a new cmdline option?

klag:
yes we should revise this aspect

mwallnoefer:
PR #606.

PR #606, opened 2025-04-14
Enhancement/jolie2openapi

klag:
I added a package, called openapi, where it is possible to convert the interfaces exposed into an input port into an openapi definition (both 2 and 3)

At the moment, the command jolie2openapi, it is not integrated with this package, because I need to refactor also jester and jolier before.

Anyway, the package can be used and should be considered the new standard way for getting openapi definitions of a jolie inputport


mwallnoefer:
Have you tried to address issue #320?

mwallnoefer:
@klag I guess your changes have broken something in joliedoc: https://github.com/jolie/jolie/actions/runs/14772313140/job/41474313768

PR #613, opened 2025-06-04
Fix extend  typelink linked type is null

Niels-Erik:
A bug exists in using a type from another file in an extender.
I created two tests that tests for this when using embed with as, and using embed without as.
The bug only seems to happen when embedding without as.

What happens is: that the Type Link is build, but the link is not added. This also gives a null exception when trying to throw new UnsupportedOperationException, so some Null checking has been put in.

I have also found the source that has the problem and applied a quickfix.

This bug has been introduced in Jolie 1.13, and did not exist in Jolie 1.12.

### DIFF
--- a/jolie/src/main/java/jolie/OOITBuilder.java
+++ b/jolie/src/main/java/jolie/OOITBuilder.java
@@ -1688,6 +1688,7 @@ public class OOITBuilder implements UnitOLVisitor {
 				extendTypes[ i ] = typeMap.get( extendedTypeDefinition );
 			} else {
 				extendTypes[ i ] = buildType( extendedTypeDefinition );
+				resolveTypeLinks(); // quickfix for bug when it is a TypeLink being build
 			}
 			result = Type.extend( result, extendTypes[ i ] );
 			i++;
--- a/jolie/src/main/java/jolie/OOITBuilder.java
+++ b/jolie/src/main/java/jolie/OOITBuilder.java
@@ -1688,8 +1688,8 @@ public class OOITBuilder implements UnitOLVisitor {
 				extendTypes[ i ] = typeMap.get( extendedTypeDefinition );
 			} else {
 				extendTypes[ i ] = buildType( extendedTypeDefinition );
-				resolveTypeLinks(); // quickfix for bug when it is a TypeLink being build
 			}
+			resolveTypeLinks(); // quickfix for bug when it is a TypeLink being build
 			result = Type.extend( result, extendTypes[ i ] );
 			i++;
 		}
--- a/jolie/src/main/java/jolie/runtime/typing/Type.java
+++ b/jolie/src/main/java/jolie/runtime/typing/Type.java
@@ -381,8 +381,10 @@ public abstract class Type {
 		} else if( t2 instanceof TypeLink ) {
 			returnType = extend( t1, (TypeLink) t2 );
 		} else {
-			throw new UnsupportedOperationException( "extension not supported between " + t1.getClass().getSimpleName()
-				+ " and " + t2.getClass().getSimpleName() );
+			String t1Name = t1 != null ? t1.getClass().getSimpleName() : "(null error)";
+			String t2Name = t2 != null ? t2.getClass().getSimpleName() : "(null error)";
+			throw new UnsupportedOperationException( "extension not supported between " + t1Name
+				+ " and " + t2Name );
 		}
 		return returnType;
 	}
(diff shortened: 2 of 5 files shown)


jolie/src/main/java/jolie/runtime/typing/Type.java
```diff
@@ -381,8 +381,10 @@ public static Type extend( Type t1, Type t2 )
 		} else if( t2 instanceof TypeLink ) {
 			returnType = extend( t1, (TypeLink) t2 );
 		} else {
-			throw new UnsupportedOperationException( "extension not supported between " + t1.getClass().getSimpleName()
-				+ " and " + t2.getClass().getSimpleName() );
+			String t1Name = t1 != null ? t1.getClass().getSimpleName() : "(null error)";
+			String t2Name = t2 != null ? t2.getClass().getSimpleName() : "(null error)";
+			throw new UnsupportedOperationException( "extension not supported between " + t1Name
+				+ " and " + t2Name );
```
mwallnoefer, line 387: It is not better to use `String.format` here?
```
throw new UnsupportedOperationException(String.format("extension not supported between %s and %s", t1Name, t2Name));
```

jolie/src/main/java/jolie/runtime/typing/Type.java
```diff
@@ -396,22 +398,24 @@ private static Type extend( Type t1, TypeImpl t2 ) {
...
 		}
 		return returnType;
 	}
 
 	private static Type extend( TypeLink t1, TypeImpl t2 ) {
-		return extend( t1.linkedType, t2 );
+		return extend( t1.linkedType, t2 ); // maybe throw exception here if linkedType is null?
```
mwallnoefer, line 410: I guess we should decide on the best place for doing these non-null checks on missing type metadata. According to me this should happen as early as possible. Does Java provide non-nullable reference types as it is the case for C#?

jolie/src/main/java/jolie/OOITBuilder.java
```diff
@@ -1689,6 +1689,7 @@ private Type getOrBuildExtendedType( TypeDefinition typeDefinition, TypeDefiniti
 			} else {
 				extendTypes[ i ] = buildType( extendedTypeDefinition );
 			}
+			resolveTypeLinks(); // quickfix for bug when it is a TypeLink being build
```
kicito, line 1692: I was thinking about a proper fix, but it requires reconsidering the build system. This adds complexity to the method, but won't affect much on performance.

Regarding this method, we can reuse `getOrBuildType` method, and perhaps add  a check, only to make it clear. So something like following?

```
	private Type getOrBuildExtendedType( TypeDefinition typeDefinition, TypeDefinition... extendedTypeDefinitions ) {
		Type result;
		extendedTypeDefinitions = Arrays.asList( extendedTypeDefinitions ).stream().filter( Objects::nonNull )
			.toArray( TypeDefinition[]::new );
		Type[] extendTypes = new Type[ extendedTypeDefinitions.length ];
		result = getOrBuildType( typeDefinition );
		int i = 0;
		if( result instanceof TypeLink ) {
			resolveTypeLinks();
		}
		for( TypeDefinition extendedTypeDefinition : extendedTypeDefinitions ) {
			extendTypes[ i ] = getOrBuildType( extendedTypeDefinition );
			if( extendTypes[ i ] instanceof TypeLink ) {
				resolveTypeLinks();
			}
			result = Type.extend( result, extendTypes[ i ] );
			i++;
		}
		return result;
	}
```

mwallnoefer: Refactored code:

```
	private Type getOrBuildExtendedType( TypeDefinition typeDefinition, TypeDefinition... extendedTypeDefinitions ) {
		Type result = getOrBuildType( typeDefinition );
		if( result instanceof TypeLink ) {
			resolveTypeLinks();
		}

		extendedTypeDefinitions = Arrays.asList( extendedTypeDefinitions ).stream().filter( Objects::nonNull )
			.toArray( TypeDefinition[]::new );
		for( TypeDefinition extendedTypeDefinition : extendedTypeDefinitions ) {
			Type extendType = getOrBuildType( extendedTypeDefinition );
			if( extendType instanceof TypeLink ) {
				resolveTypeLinks();
			}
			result = Type.extend( result, extendType );
		}

		return result;
	}
```

jolie/src/main/java/jolie/runtime/typing/Type.java
mwallnoefer: > Another idea could be to create a getter for `linkedType` that throws a `TypeLinkNotResolvedException` or a null exception, to show immediately what the problems is. Ideas?

Ok. But how to guarantee that everybody (also in the future) uses the new getter instead of accessing the field directly? I mean the `private` visibility level does not help much since we are always acting in the `Type` class.

Maybe @fmontesi as the original author would like to add something?

jolie/src/main/java/jolie/OOITBuilder.java
Niels-Erik: I've replaced the code with the refactored code.

jolie/src/main/java/jolie/runtime/typing/Type.java
kicito: > Java does not provide non-nullable reference types as C# does. Java do however have the Optional type, which could be one way to solve it, though it could require a lot of code change. I do not like the use of `assert` as it is not used by default, and I also don't see how it would've improved the situation, but that might be my inexperience, where would the `assert` be and how would it help?

Sorry I didn’t elaborate clearly earlier. I was commenting based on the added comment in the code.  I don’t think we need to change anything here. When running a .ol file, all TypeLinks are expected to be resolved by having .linkType assigned at this stage. Any errors related to TypeLink should be raised earlier in the stages (e.g., in the Parser or SymbolResolver). Since there’s nothing a developer can do in the Jolie file to fix the null pointer here, I’d discourage adding an exception.

To improve maintainability of this method, adding an assert along with a comment may be the way to go. At least it will help future developers understand the method more easily when they come across it.

kicito (approved):

fmontesi:
Thank you all for the discussion, this looks pretty good! :)

jolie/src/main/java/jolie/runtime/typing/Type.java
Niels-Erik: I've added the assertions now.

PR #615, opened 2025-06-11
Escape control characters on value printing on terminal

kicito:
Introduces a feature to escape ISO control characters in console output. Force escape for logger, ValuePrettyPrinter, and tracer methods; add escaping logic in the ConsoleService.

### DIFF
--- a/javaServices/coreJavaServices/src/main/java/joliex/io/ConsoleService.java
+++ b/javaServices/coreJavaServices/src/main/java/joliex/io/ConsoleService.java
@@ -31,18 +31,18 @@ import java.text.SimpleDateFormat;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.Map;
-
 import jolie.runtime.JavaService;
 import jolie.runtime.Value;
 import jolie.runtime.embedding.RequestResponse;
+import jolie.runtime.typing.TypeCastingException;
 import jolie.util.Helpers;
 
 public class ConsoleService extends JavaService {
 	private Map< String, String > sessionTokens;
 	private boolean sessionListeners = false;
 	private boolean enableTimestamp = false;
-	private boolean escapeControlChars = false;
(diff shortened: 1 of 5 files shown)


fmontesi:
@mwallnoefer @klag @kicito This looks good but I'm a bit in doubt about introducing a stateful configuration method (`escapeControlChars`). I just wanna point out that this is significant, but I don't see a good alternative, also because `println`/`print` can't get arguments.

We might wanna make this more explicit though, e.g., call the operation `config` and make `escapeControlChars` a parameter of `config` (for now the only one) or at least call the operation `setEscapeControlChars`. What do you think?

kicito:
There was a case where the API response included special characters that caused the terminal to misbehave when Jolie printed or dumped the response to the console. This PR introduces a workaround to handle such cases.

If we force escaping the special character, programs that rely on std out/in for communication might be affected by this change. It is rare nowadays (programs rely on special characters), but it is worth noting this potential side effect.

@fmontesi How about Jolie prints to stderr when encountering a special character, and it is the developer's task to handle the issue themselves (or with `config` operation as suggested)? IMO, we are dealing with an edge case, noting much Jolie should do about it.

fmontesi:
It could break existing user code. The current behaviour of println is that
it accepts any tree and prints its root value. The subnodes are just
ignored. I suspect this feature is used extensively in existing code.

klag:
mmmm this is a technical debt on the declaration of println and print. They should be:

println( string )( void ) and not println( undefined )( void )

So, we could leave the undefined as a type, but adding a subfield escapeSpecialCharacter.
We can put the expected type as a comment and add a check on the print invocations, printong out a warning like:

WARNING: the undefined request type for println and print is deprecated. The expected type now is:
type PrintRequest: string {
.escapeSpecialCharacters: bool // default true
}

From version XX  undefined will be replaced with the new one.

PR #621, opened 2025-08-05
added findAll to StringUtils

klag:
I added findAll in order to find all the occurencies of a string into a target string.
the existing api find, just finds only the first occurence


fmontesi:
To understand: I guess you could use `match` and a regex with a star (`*`) to do the same thing, right?

klag:
match is usually exploited for just matching a string into another

fmontesi:
Ah, I see, the key difference is that in a `match` I have to use capturing groups in the request, whereas here capturing (to return the results) is always done implicitly, I just have to pass a normal regex. Correct?
If so, I think we're good to go.

Issue #367, opened 2021-11-23
openapi2jolie - missing generating an operation signature


SuperMauro1966:
**Describe the bug**
Trying to generate a jolie client from the openapi file included below,
the type SelectedSession belonging to createPDUSession is not generated.

**To Reproduce**
openapi2jolie DanieleRossi-UEDigitalEntityNBI_API-1.0.0-resolved.json UEDIgitalEntityNBI_API . http

**Expected behavior**
A full list of type generated from the corresponding json conterpart,
a correct signature for the operations

**Screenshots**
---

**Desktop (please complete the following information):**
 - OS: Linux (Ubuntu 18.04)
 - Jolie version 1.10.5
 - Java Version jdk 11

**Additional context*
[DanieleRossi-UEDigitalEntityNBI_API-1.0.0-resolved.json.txt](https://github.com/jolie/jolie/files/7587094/DanieleRossi-UEDigitalEntityNBI_API-1.0.0-resolved.json.txt)
*

mwallnoefer:
The problem here is that _openapi2jolie_ supports custom schema attributes under the `#/definitions/` section, but the JSON in the attachment provides them under `#/components/schemas/`.

PR #623, opened 2025-08-21
Enhancement/jolie2openapi tool

klag:
Refactoring of jolie2openapi tool

added also package args.ol for managing args for a jolie process

during tests some errors appears because of multiple instance of openapi installed by the jolie2openapi test, let us discuss how to avoid them. Anyway tests pass

### DIFF
--- a/packages/args.ol
+++ b/packages/args.ol
@@ -2,18 +2,18 @@ from string-utils import StringUtils
 from console import Console
 
 
-type GetParsedArgsRequest {
+type ParseArgsRequest {
     args*: string 
-    argsMap*: void {
-        name: string        // name of the expected argument
-        optional: bool      // specifies if the argument is optional
-        description: string // describes the argument
-        isFlag: bool       // if true, the argument is a flag (e.g. --verbose)
+    argsMap* {
+        name: string        //< name of the expected argument
+        optional: bool      //< specifies if the argument is optional
+        description: string //< describes the argument
+        isFlag: bool        //< if true, the argument is a flag (e.g. --verbose)
     }
 
 }
 
-type GetParsedResponse {
+type ParseArgsResponse {
     parsedArgs* {
         name: string 
         value: string | void
@@ -22,7 +22,7 @@ type GetParsedResponse {
 
 interface ArgsInterface {
     RequestResponse:
-        getParsedArgs( GetParsedArgsRequest )( GetParsedResponse ) throws ParametersError( string ) Help
+        parseArgs( ParseArgsRequest )( ParseArgsResponse ) throws ParsingError( string ) Help
 
 }
 
@@ -58,12 +58,12 @@ service Args {
 
 
     init {
-        install( ParametersError => nullProcess )
+        install( ParsingError => nullProcess )
         install( Help => nullProcess )
     }
 
     main {
-        getParsedArgs( request )( response ) {
+        parseArgs( request )( response ) {
             // compose all the args into a single string, this allows for having strings as arguments
             args_string = ""
             for( a in request.args ) {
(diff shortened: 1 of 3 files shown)


mwallnoefer:
@klag: sounds good, but a branch re-base simplifies the review (e.g. `findAll()` has already been integrated, so shouldn't appear anymore).

PS Did you per chance manage to address some limitations listed in issue #367 (not easy to fix, I know)?

klag:
> @klag: sounds good, but a branch re-base simplifies the review (e.g. `findAll()` has already been integrated, so shouldn't appear anymore).
> 
> PS Did you per chance manage to address some limitations listed in issue #367 (not easy to fix, I know)?

rebase done

klag:
@mwallnoefer as far as issue https://github.com/jolie/jolie/issues/367 is concerned, so far I did not refactor the way: openapi -> jolie, but only jolie -> openapi

packages/args.ol
```diff
@@ -0,0 +1,115 @@
+from string-utils import StringUtils
+from console import Console
+
+
+type GetParsedArgsRequest {
+    args*: string 
+    argsMap*: void {
+        name: string        // name of the expected argument
```
fmontesi, line 8: You need //< in these.

packages/args.ol
```diff
@@ -0,0 +1,115 @@
...
+type GetParsedResponse {
+    parsedArgs* {
+        name: string 
+        value: string | void
+    }
+}
+
+interface ArgsInterface {
+    RequestResponse:
+        getParsedArgs( GetParsedArgsRequest )( GetParsedResponse ) throws ParametersError( string ) Help
```
fmontesi, line 25: I'd just call this `parseArgs`, and the types ParseArgsRequest and ParseArgsResponse.

packages/args.ol
```diff
@@ -0,0 +1,115 @@
+from string-utils import StringUtils
+from console import Console
+
+
+type GetParsedArgsRequest {
+    args*: string 
+    argsMap*: void {
```
fmontesi, line 7: Is there something more semantic we could call this? `argTypes`, `argsSpec`, `schema`, ...?

fmontesi, line 7: Don't need the void.

packages/args.ol
fmontesi, line 25: Why is Help a fault? Are we forcing everyone to use --help as escape hatch? I'd just return the response normally and then leave it to the client to check if --help was specified, it looks like non-exceptional logic.

fmontesi, line 25: `ParametersError` is a bit surprising since parameters hasn't been mentioned before, maybe `ParsingError`?

fmontesi (changes requested):

fmontesi:
@klag what do you mean by multiple installations of openapi?

packages/args.ol
klag: Because args has been thouth for being used from console. Usually when you type --help the command prints out the help list and then exit.

fmontesi:
I have one last comment, but I'm just gonna leave it here for the future as I think this is good to merge in the meantime.

It is weird that `args` is not functional, in the sense that in my opinion it should just parse arguments and return a result or a parse error. It shouldn't print things on screen or have other side-effects like that, that should be the job of another service that uses `args` as a dependency. Maybe we wanna make this split in the future.

mwallnoefer:
@klag I guess we are getting some unit test problems here, could you please tryto reproduce (https://github.com/jolie/jolie/actions/runs/17378372872/job/49329820985)?
```
jolie2openapi.ol...                 Creating openapi file...
Sep 01, 2025 1:01:44 PM OpenApiUtils
SEVERE: jolie.InterpreterException: /home/runner/work/jolie/jolie/test/../packages/openapi.ol:216: Address already in use
	at jolie.OOITBuilder.error(OOITBuilder.java:322)
	at jolie.OOITBuilder.visit(OOITBuilder.java:732)
	at jolie.lang.parse.UnitOLVisitor.visit(UnitOLVisitor.java:429)
	at jolie.lang.parse.UnitOLVisitor.visit(UnitOLVisitor.java:92)
	at jolie.lang.parse.ast.InputPortInfo.accept(InputPortInfo.java:85)
	at jolie.lang.parse.ast.OLSyntaxNode.accept(OLSyntaxNode.java:47)
	at jolie.OOITBuilder.visit(OOITBuilder.java:795)
	at jolie.OOITBuilder.build(OOITBuilder.java:379)
	at jolie.Interpreter.buildOOIT(Interpreter.java:1227)
	at jolie.Interpreter.init(Interpreter.java:988)
	at jolie.Interpreter$StarterThread.run(Interpreter.java:1113)

jolie2openapi.ol...                 Creating openapi file... 
...
```

Unfortunately our test framework did not cover this.

mwallnoefer:
Also the failing assertions in the new _args_ library seem to get ignored:
```
args.ol...                          ❌ Error: Missing required argument: arg4

Documentation:

--arg1                     [required] [require value] This is the first argument
--arg2                     [required] [require value] This is the second argument
--arg3                     [required] [require value] This is the third argument
--arg4                     [required] [require value] This is an optional argument
❌ Error: Argument arg4 is a flag and cannot have a value


Documentation:

--arg1                     [required] [require value] This is the first argument
--arg2                     [required] [flag argument] This is the second argument
--arg3                     [required] [require value] This is the third argument
--arg4                     [optional] [flag argument] This is an optional argument
passed ✅
```

I have managed to fix the outputs, but not the cause: https://github.com/jolie/jolie/pull/643/commits/c1df27e2f26d04d604c3cca913d95cdf01580213

Issue #552, opened 2025-01-08
import exists poorly on '.' as path/file


Niels-Erik:
**Describe the bug**
If an import only has a dot,`.`, for the file path, Jolie will crash with the error: `fromIndex(0) > toIndex(-1)`

**To Reproduce**
1. Create `bug.ol` with `from . import anything`
2. Run `jolie bug.ol`
3. See error

**Expected behavior**
Jolie should tell that I the path/file does not exist or that I wrote the path wrong.

**Desktop (please complete the following information):**
 - OS: windos
 - Jolie version 1.13.0-git
 - Java Version 21

mwallnoefer:
Could you please try to apply the fix of PR #551?

Niels-Erik:
> Could you please try to apply the fix of PR #551?

Same issue

kicito:
Thanks for reporting this, #555 should fix it.

Issue #528, opened 2024-09-30
Error in jolie for windows


adelmogentilini:
The Windows version of JOLIE, whether installed by downloading the binary or built locally with `mvn clean` install, gives me the error "*Cannot invoke "jolie.State.root()" because the return value of "jolie.ExecutionThread.getState()" is null*" when running any .ol file.

**To Reproduce**
git clone https://github.com/jolie/jolie.git
cd jolie
mvn install

set uo enviroment variable JOLIE_HOME to directory and copy dist in this directory and the launcher in bin subdir. Add the bin subdir to path.
jolie -h is correct
jolie file.ol 
==  > Cannot invoke "jolie.State.root()" because the return value of "jolie.ExecutionThread.getState()

**Expected behavior**
I expect run of file.ol

**Desktop (please complete the following information):**
 - OS: Windows
 - Jolie version master tree
 - Java Version 21

mwallnoefer:
Jolie 1.12.2 with Java 21 on Windows 10 works, so it has to do with the latest changes done in git.

fmontesi:
Is this bug still there?

Niels-Erik:
I've just gotten this "bug" on my desktop pc today.
However I think it comes after I was cleaning my C drive to get more space, since it worked before that.
I have tried to reinstall java jdk 21 and jolie 1.12.2.

Niels-Erik:
I forgot that I cannot run Jolie files that is in any subfolder of my user, since there is a space in my name, Niels Erik...
However I have found a way to reproduce the bug, by trying to the following code:
```
interface addInterface {
    RequestResponse: 
    add(int)(int)
}

service bug {

    inputPort httpPort {
        Location: "socket://localhost:8080" 
        protocol: http {.format = json}
        interfaces: addInterface
    }

    main{
        
        [add(req)(res){
            res = req++
        }]
    }
}
```
This would run if you delete the `{.format = json}` part.
The stacktrace from running with --stacktrace is:
```
java.lang.NullPointerException: Cannot invoke "jolie.State.root()" because the return value of "jolie.ExecutionThread.getState()" is null
        at jolie.runtime.VariablePath.getRootValue(VariablePath.java:129)
        at jolie.runtime.VariablePath.getValueOrNull(VariablePath.java:243)
        at jolie.runtime.VariablePath.evaluate(VariablePath.java:444)
        at jolie.runtime.expression.InlineTreeExpression$AssignmentOperation.run(InlineTreeExpression.java:90)
[...]

PR #626, opened 2025-09-02
Added info for unexpected errors

Niels-Erik:
It is useful to inform the user when unexpected bugs happen, like #552 and #528/#625

### DIFF
--- a/jolie-cli/src/main/java/jolie/Jolie.java
+++ b/jolie-cli/src/main/java/jolie/Jolie.java
@@ -100,6 +100,7 @@ public class Jolie {
 			}
 			exitCode = 3;
 		} catch( Exception e ) {
+			System.err.println("Unexpected error, run with --stackTraces and report on github:");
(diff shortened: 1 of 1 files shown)


jolie-cli/src/main/java/jolie/Jolie.java
```diff
@@ -100,6 +100,7 @@ public static void main( String[] args ) {
 			}
 			exitCode = 3;
 		} catch( Exception e ) {
+			System.err.println("Unexpected error, run with --stackTraces and report on github:");
```
fmontesi, line 103: We can be a bit more informative, I suggest:

```
System.err.println( "An unexpected error occurred. If you think this is a bug, you can report it on GitHub at https://github.com/jolie/jolie/issues" );
if( !printStackTraces ) {
    System.err.println( "You can run Jolie with --stackTraces for more information." );
}
```

fmontesi (changes requested):

Issue #625, opened 2025-09-02
variables cannot be resolved before init


Niels-Erik:
Basically the issue from #528, now as its own issue.

Error can be found the following way:
> However I have found a way to reproduce the bug, by trying to the following code:
> ```
> interface addInterface {
>     RequestResponse: 
>     add(int)(int)
> }
> 
> service bug {
> 
>     inputPort httpPort {
>         Location: "socket://localhost:8080" 
>         protocol: http {.format = json}
>         interfaces: addInterface
>     }
> 
>     main{
> 
>         [add(req)(res){
>             res = req++
>         }]
> 
>     }
> 
> }
> ```
> 
> This would run if you delete the `{.format = json}` part.
> 
> The stacktrace from running with --stacktrace is:
> 
> ```
> java.lang.NullPointerException: Cannot invoke "jolie.State.root()" because the return value of "jolie.ExecutionThread.getState()" is null
>         at jolie.runtime.VariablePath.getRootValue(VariablePath.java:129)
>         at jolie.runtime.VariablePath.getValueOrNull(VariablePath.java:243)
>         at jolie.runtime.VariablePath.evaluate(VariablePath.java:444)
>         at jolie.runtime.expression.InlineTreeExpression$AssignmentOperation.run(InlineTreeExpression.java:90)
>         at jolie.runtime.expression.InlineTreeExpression.evaluate(InlineTreeExpression.java:143)
>         at jolie.OOITBuilder.visit(OOITBuilder.java:509)
>         at jolie.lang.parse.UnitOLVisitor.visit(UnitOLVisitor.java:429)
>         at jolie.lang.parse.UnitOLVisitor.visit(UnitOLVisitor.java:92)
>         at jolie.lang.parse.ast.InputPortInfo.accept(InputPortInfo.java:85)
>         at jolie.lang.parse.ast.OLSyntaxNode.accept(OLSyntaxNode.java:47)
>         at jolie.OOITBuilder.visit(OOITBuilder.java:615)
>         at jolie.OOITBuilder.build(OOITBuilder.java:216)
>         at jolie.Interpreter.buildOOIT(Interpreter.java:1239)
>         at jolie.Interpreter.init(Interpreter.java:996)
>         at jolie.Interpreter.run(Interpreter.java:1079)
>         at jolie.Jolie.main(Jolie.java:86)
> ``` 

 _Originally posted by @Niels-Erik in [#528](https://github.com/jolie/jolie/issues/528#issuecomment-2506783726)_

[...]

PR #647, opened 2025-12-21
VariablePath: introduce a fake fallback value for a yet unknown init state

mwallnoefer:
This allows to read it for the initalisation phase in order to address issue #625


mwallnoefer:
According to me this `format = json` (vs. the correct `format = "json"`) is a parsing issue, which should better be denied and not silently ignored.

Otherwise the patch should work and we are able to remove this dirty `getState()` kludge. So I am very happy to see it merged.

PR #650, opened 2026-02-04
make langue server on windows with spaces in usernames

Niels-Erik:
This change improves on Jolie being able to handle spaces in paths, focusing on making the language server work on windows, when the username have spaces in it.
(ps. I cannot edit the title)

Fixed not being able to load jar files via loadLibrary@runtime or jolie files via loadEmbeddedService@runtime in folders with spaces.

This makes the language server able to run on windows whose username contains spaces, if https://github.com/jolie/languageserver/pull/6 gets merged.


fmontesi:
LGTM. @kicito and @mwallnoefer can anybody have a double look at this? It's a sensitive code path.

jolie/src/main/java/jolie/JolieClassLoader.java
```diff
@@ -344,8 +344,7 @@ public void addJarResource( String jarName )
 			// Try again with URI method, since from
 			// Java 17's URLClassLoader.findResource fixes undocumented
```
mwallnoefer: Seems ok, but the comment would need to be adjusted since `URLClassLoader.findResource` gets removed with your change.

PR #652, opened 2026-02-08
new with syntax as stated in #196

### LINKED ISSUE #196: bug in with primitive
see test in branch **with_bug**: https://github.com/jolie/jolie/tree/with_bug

Niels-Erik:
This is my attempt to fix #196 where `..` is a new` Scanner.TokenType`.

Main changes are:
- update all files that used the with statement
- Scanner integrates the new TokenType
- Parser updated to use new TokenType for with statements
- Parser sees the new `..` as two `.` for import statements

### DIFF
--- a/include/database.iol
+++ b/include/database.iol
@@ -63,11 +63,11 @@ RequestResponse:
 	 *
 	 * Example with HSQLDB:
 	 * with ( connectionInfo ) {
-	 *     .username = "sa";
-	 *     .password = "";
-	 *     .host = "";
-	 *     .database = "file:weatherdb/weatherdb"; // "." for memory-only
-	 *     .driver = "hsqldb_embedded"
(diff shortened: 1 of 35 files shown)


fmontesi:
(Sorry, forgot to put this comment sooner.) @klag, how does this look to you? I remember you're one of the biggest users of this feature.

kicito (commented):
Will this break backward compatibility with the previous version? Would it be better to allow both `.` and `..` for now and add a deprecation warning for the single dot usage?

Niels-Erik:
> Will this break backward compatibility with the previous version? Would it be better to allow both `.` and `..` for now and add a deprecation warning for the single dot usage?

It properly will break backwards compatibility.
I'll have to look into the code again to make sure and see how difficult it is to change to keep both,

